Issue #42, opened 2019-03-15
Grants controller necessary permissions through policy file


Jeffwan:
As a customer, I'd like to use policy file and grant permission to node groups rather than using `aws_access_key_id` and `aws_secret_access_key`.  It's hard for me to provide fine grain control for it because documentation doesn't state minimum permissions CSI driver needs and I have to provide admin access to avoid permission errors and failures. 

Solution: 
I would suggest to explicitly list permissions in a policy file and then attach to node group. 

/feature

christopherhein:
Once we have the permissions necessary we can add it to `eksctl` like the EBS CSI - https://github.com/weaveworks/eksctl/pull/652

PR #62, opened 2019-04-07
Add iam policy for FSx driver

Jeffwan:
Improve documentation on permission setups. 

I have use case that need to attach policy to node groups rather than using credential. 

**What testing is done?** 
Manual testing

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -54,6 +54,30 @@ kubectl apply -f secret.yaml
 
 * Using worker node instance profile - grant all the worker nodes with proper permission by attach policy to the instance profile of the worker.
 
+```json
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -54,6 +54,40 @@ kubectl apply -f secret.yaml
...
+    {
+      "Effect": "Allow",
+      "Action": [
+        "fsx:*"
+      ],
+      "Resource": ["*"]
+    },
+    {
+      "Effect": "Allow",
+      "Action": "s3:*",
```
leakingtapan: I think `s3:*` and `ec2:CreateNetworkInterface` is not needed by driver itself

Jeffwan: I see. I remembered I get some Access Denied error. Let me double check on my end.

leakingtapan:
Fixes: #42

docs/README.md
Jeffwan: @leakingtapan Make the change and remove unnecessary permissions. Sorry for late update

PR #69, opened 2019-07-20
Update to CSI v1.1.0

wongma7:
**Is this a bug fix or adding new feature?** new feature

**What is this PR about? / Why do we need it?** fixes https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/39
- NodePublish request VolumeAttributes -> VolumeContext
- ValidateVolumeCapabilitiesResponse is different
- ControllerExpandVolume, NodeGetVolumeStats, and NodeExpandVolume are unimplemented.

**What testing is done?** ~WIP: it builds...however, tests still need to be updated. And ValidateVolumeCapabilities also needs to be implemented fully.~
All tests pass.

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -20,7 +20,7 @@ import (
 	"context"
 	"strings"
 
-	csi "github.com/container-storage-interface/spec/lib/go/csi/v0"
+	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/kubernetes-sigs/aws-fsx-csi-driver/pkg/cloud"
 	"github.com/kubernetes-sigs/aws-fsx-csi-driver/pkg/util"
 	"google.golang.org/grpc/codes"
@@ -93,7 +93,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Errorf(codes.Internal, "Filesystem is not ready: %v", err)
 	}
 
-	return newCreatVolumeResponse(fs), nil
+	return newCreateVolumeResponse(fs), nil
 }
 
 func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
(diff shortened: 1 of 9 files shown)


pkg/driver/controller.go
```diff
@@ -166,10 +166,20 @@ func (d *Driver) ValidateVolumeCapabilities(ctx context.Context, req *csi.Valida
...
-		Supported: supported,
-	}, nil
+	confirmed := d.isValidVolumeCapabilities(volCaps)
+	if confirmed {
+		return &csi.ValidateVolumeCapabilitiesResponse{
+			Confirmed: &csi.ValidateVolumeCapabilitiesResponse_Confirmed{
+				// TODO if volume context is provided, should validate it too
+				// VolumeContext:      req.GetVolumeContext(),
+				VolumeCapabilities: volCaps,
+				// TODO if parameters are provided, should validate them too
```
wongma7: I have left 2 TODOs here for a follow-up PR. CreateVolume could use the same validation

tests/sanity/sanity_test.go
```diff
@@ -53,6 +53,7 @@ var _ = AfterSuite(func() {
 })
 
 var _ = Describe("AWS FSx for Lustre CSI Driver", func() {
+	_ = os.MkdirAll("/tmp/csi", os.ModePerm)
```
leakingtapan: Why is this needed?

wongma7: sanity used to MkdirAll the paths for us, but now it only does Mkdir causing tests to fail without this (locally and in travis): https://github.com/kubernetes-csi/csi-test/commit/267dbbcb874cf12332ec966f372a10ad8df246d6#diff-cdf6b6b472a77c21dd6f192b6f3241eaL262

onur-sam-gtn-ai:
README is not updated and this change broke all our EKS clusters relying on FSx suddenly.
Master version is no longer CSI 0.3. Our manifests were pointing to the latest tag and that broke the driver overnight for us.

wongma7:
@onur-sam-gtn-ai sorry about that, I've opened https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/76 to update the README. Please use the v0.1.0 tag as the latest tag is unstable, especially while the driver is alpha. (Beta is coming soon). Thanks.

PR #70, opened 2019-07-22
 Add volume context and parameters validation to ValidateVolumeCapabilities

wongma7:
**Is this a bug fix or adding new feature?** bug fix

**What is this PR about? / Why do we need it?** CSI 1.0 changed ValidateVolumeCapabilities, now it should return all the fields it has validated instead of just `Supported`.  So, add volume context and parameters validation to the function just like how there is already volume capabilities validation

This is a follow-up to https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/69 , https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/69 should be reviewed first
**What testing is done?** `make test` passes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ require (
 	github.com/container-storage-interface/spec v1.1.0
 	github.com/davecgh/go-spew v1.1.0 // indirect
 	github.com/golang/mock v1.2.0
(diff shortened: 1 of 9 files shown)


wongma7:
This patch is not essential. I'm not even sure in what situations ValidateVolumeCapabilities will be called by kubernetes. It could be a bugfix should someone run into an issue with it.

Issue #65, opened 2019-05-30
flock mountoption


tsigalko2003:
We are running some bioinformatic pipeline with NextFlow, which requires file lock when creating the local database to manage the subprocesses.  It requires mountoptions flock, or localflock enabled. Unfortunately, I got the error message saying the mountoption is not 
supported.

leakingtapan:
Related to #30

tsigalko2003:
Totally understand the volume of work it may need. Really appreciate if we can get flock/flocallock as your earliest convenience.  Nextflow can not work without file lock at the current stage.

Issue #30, opened 2019-02-27
Mount Option Support


leakingtapan:
There are several case to specify mount options for lustre fs:
* http://wiki.lustre.org/Mounting_a_Lustre_File_System_on_Client_Nodes

PR #74, opened 2019-07-24
Implement mount options support

wongma7:
**Is this a bug fix or adding new feature?** new feature

**What is this PR about? / Why do we need it?** fix https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/65 and fix https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/30. Include the flock option in our example StorageClasses because according to http://wiki.lustre.org/Mounting_a_Lustre_File_System_on_Client_Nodes it's enabled by default and has no performance impact.

**What testing is done?** make test passes. ~will do an actual test with one of the examples and update.~ done, works

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -39,7 +39,8 @@ Following sections are Kubernetes specific. If you are Kubernetes user, use foll
 
 ### Features
 * Static provisioning - FSx for Lustre file system needs to be created manually first, then it could be mounted inside container as a volume using the Driver.
-* Dynamic provisioning - uses persistence volume claim (PVC) to let the Kuberenetes to create the FSx for Lustre filesystem for you and consumes the volume from inside container.
+* Dynamic provisioning - uses persistent volume claim (PVC) to let the Kuberenetes to create the FSx for Lustre filesystem for you and consumes the volume from inside container.
+* Mount options - mount options can be specified in persistent volume (PV) to define how the volume should be mounted.
 
 ### Installation
 #### Set up driver permission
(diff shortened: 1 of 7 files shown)


examples/kubernetes/dynamic_provisioning/specs/storageclass.yaml
```diff
@@ -6,3 +6,5 @@ provisioner: fsx.csi.aws.com
 parameters:
   subnetId: subnet-056da83524edbe641
   securityGroupIds: sg-086f61ea73388fb6b
+mountOptions:
+  - flock
```
leakingtapan: I think if `flock` is only required under certain situation. Could you create a new `MountOption` example for just this use case? I suspect user most of the time just copy paste examples and assume it as best practices, which might be not true for using `flock` option in dynamic provisioning generally.

Same comment in other examples.

wongma7: I think flock a reasonable default/best practice because we put it in our docs as the default and tell users to remove it if they don't want it: 
"flock – mounts your file system with file locking enabled. If you do not want file locking enable, use the command without flock." https://docs.aws.amazon.com/fsx/latest/LustreGuide/mounting-ec2-instance.html. So in the same way, I think it's up to the user to remove flock from the StorageClass if they don't want it.

Like, there is nothing preventing a user from (inadvertently) mounting the same PVC in different pods (maybe they scale up their deployment) so flock seems to be the "safe" option

docs/README.md
```diff
@@ -39,7 +39,8 @@ Following sections are Kubernetes specific. If you are Kubernetes user, use foll
 
 ### Features
 * Static provisioning - FSx for Lustre file system needs to be created manually first, then it could be mounted inside container as a volume using the Driver.
-* Dynamic provisioning - uses persistence volume claim (PVC) to let the Kuberenetes to create the FSx for Lustre filesystem for you and consumes the volume from inside container.
+* Dynamic provisioning - uses persistent volume claim (PVC) to let the Kuberenetes to create the FSx for Lustre filesystem for you and consumes the volume from inside container.
+* Mount options - mount options can be specified in persistent volume (PV) to define how the volume should be mounted.
```
leakingtapan:
```suggestion
* Mount options - mount options can be specified in storageclass to define how the volume should be mounted.
```

PR #82, opened 2019-08-14
Fix manifest csi driver object

leakingtapan:
Fix manifest csi driver object

/cc @wongma7


wongma7:
/lgtm
oops, thanks

Issue #99, opened 2019-09-30
Fix IAM  policy for instance profile


leakingtapan:
**Is your feature request related to a problem? Please describe.**
Fix IAM  policy for instance profile to include S3 R/W permission

PR #100, opened 2019-10-04
Update README for IAM policy

leakingtapan:
Update README for IAM policy. This is needed for creating Fsx filesystem with s3 data repository using instance profile.

Fixes: #99

/cc @Jeffwan


Jeffwan:
/lgtm

Cool. Thanks for update. I will update policy on kubeflow side.

PR #109, opened 2019-10-21
Add example for lusture max cached size tuning

leakingtapan:
Add example for lusture max cached size tuning


examples/kubernetes/max_cache_tuning/README.md
```diff
@@ -0,0 +1,35 @@
+## Tuning Lustre Max Memory Cache
+This example shows how to set lustre `llite.*.max_cached_mb` using init container. Lustre client interacts with lustre kernal module for data caching at host level. Since the cache resides in kernal space, it won't be counted toward application container's memory limit. Sometimes it is desireable to reduce the lustre cache size to limit memory consumption at host level.
```
Perif: Typo on *kernal*

Perif: perhaps good to mention that the example below is to limit the Lustre cache size to 32MB but other values may be selected depending on what makes sense for the workload.

PR #130, opened 2020-01-16
Add support for FSx for lustre create API deployment type

chyz198:
Adding new feature
FSx for Lustre is going to support couple of new deployment types
**What testing is done?** 
1) Unit test
2) Manual end to end test including static and dynamic provisioning


examples/kubernetes/dynamic_provisioning_withdeploymenttype/README.md
```diff
@@ -0,0 +1,80 @@
...
+metadata:
+  name: fsx-sc
+provisioner: fsx.csi.aws.com
+parameters:
+  subnetId: subnet-0cd1914f0b9041ebe
+  securityGroupIds: sg-07662e27978b3b360,sg-057948b58dd38e68b
+  s3ImportPath: s3://evanzempty
+  s3ExportPath: s3://evanzempty
+  deploymentType: SCRATCH_2
+mountOptions:
```
leakingtapan: You should define the yaml list as multi-line elements instead of using comma: 
```suggestion
mountOptions:
  - flock
  - noatime
```

examples/kubernetes/dynamic_provisioning_withdeploymenttype/README.md
```diff
@@ -0,0 +1,80 @@
+## Dynamic Provisioning with DeploymentType
```
leakingtapan:
```suggestion
## Dynamic Provisioning with Deployment Type
```

examples/kubernetes/dynamic_provisioning_withdeploymenttype/README.md
```diff
@@ -0,0 +1,80 @@
...
+This example shows how to create a FSx for Lustre filesystem using persistence volume claim (PVC) with DeploymentType specified
```
leakingtapan:
```suggestion
This example shows how to create a FSx for Lustre filesystem using persistence volume claim (PVC) with Deployment Type specified
```

examples/kubernetes/dynamic_provisioning_withdeploymenttype/README.md
```diff
@@ -0,0 +1,80 @@
...
+  s3ExportPath: s3://evanzempty
+  deploymentType: SCRATCH_2
+mountOptions:
+  - flock,noatime
+```
+* subnetId - the subnet ID that the FSx for Lustre filesystem should be created inside.
+* securityGroupIds - a comman separated list of security group IDs that should be attached to the filesystem
+* s3ImportPath(Optional) - S3 data repository you want to copy from S3 to persistent volume
+* s3ExportPath(Optional) - S3 data repository you want to export new or modified files from persistent volume to S3
+* deploymentType: FSx for lustre now provides three deployment types which uses different EC2 instances types
```
leakingtapan: Add the list of three valid deployment types here?

examples/kubernetes/dynamic_provisioning_withdeploymenttype/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Deploy the Application
+Create PVC, storageclass and the pod that consumes the PV:
+```sh
+>> kubectl apply -f examples/kubernetes/dynamic_provisioning_withdeploymenttype/specs/storageclass.yaml
+>> kubectl apply -f examples/kubernetes/dynamic_provisioning_withdeploymenttype/specs/claim.yaml
+>> kubectl apply -f examples/kubernetes/dynamic_provisioning_withdeploymenttype/specs/pod.yaml
+```
+
+### Use Case 1: Access S3 files from Lustre filesystem
```
leakingtapan: the following is a copy & paster from the s3 example, since we have similar example there already, there is no need to duplicate the use cases but instead we should focus on the use case of deployment type itself here

examples/kubernetes/dynamic_provisioning_withdeploymenttype/specs/storageclass.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: fsx-sc
+provisioner: fsx.csi.aws.com
+parameters:
+  subnetId: subnet-0cd1914f0b9041ebe
+  securityGroupIds: sg-07662e27978b3b360,sg-057948b58dd38e68b
+  s3ImportPath: s3://evanzempty
+  s3ExportPath: s3://evanzempty
+  deploymentType: SCRATCH_2
+mountOptions:
+  - flock,noatime
```
leakingtapan:
```suggestion
  - flock
  - noatime
```

pkg/cloud/cloud.go
```diff
@@ -168,10 +185,16 @@ func (c *cloud) DescribeFileSystem(ctx context.Context, fileSystemId string) (*F
...
+	mountName := "fsx"
+	if fs.LustreConfiguration.MountName != nil {
+		mountName = *fs.LustreConfiguration.MountName
+	}
+
 	return &FileSystem{
 		FileSystemId: *fs.FileSystemId,
 		CapacityGiB:  *fs.StorageCapacity,
 		DnsName:      *fs.DNSName,
+		MountName:    mountName,
```
leakingtapan: Can you just set `*fs.LustreConfiguration.MountName` here?

examples/kubernetes/dynamic_provisioning_withdeploymenttype/README.md
chyz198: For deploymentType itself, there is not much to say here since from customer point of view, the usage is the same.

pkg/cloud/cloud.go
chyz198: It is better to add the nil check here just in case.

examples/kubernetes/dynamic_provisioning_withdeploymenttype/README.md
chyz198: Sure.

examples/kubernetes/dynamic_provisioning_withdeploymenttype/README.md
wongma7: if it's similar, maybe we can just edit the existing example? It's basically one new optional parameter/line in the storageclass, if it doesn't change customer usage/workflow then it should be fine

pkg/cloud/cloud.go
```diff
@@ -119,6 +122,10 @@ func (c *cloud) CreateFileSystem(ctx context.Context, volumeName string, fileSys
 		lustreConfiguration.SetExportPath(fileSystemOptions.S3ExportPath)
 	}
 
+	if fileSystemOptions.DeploymentType != "" {
+		lustreConfiguration.SetDeploymentType(fileSystemOptions.DeploymentType)
```
wongma7: should we validate the deploymenttype or let fsx API throw an error if the deploymenttype is invalid? The driver will retry forever , so it could prevent some useless api calls in case somebody makes a typo...

examples/kubernetes/dynamic_provisioning_withdeploymenttype/README.md
chyz198: Ok, I will make the change to modify the existing examples.

pkg/cloud/cloud.go
chyz198: In case of a typo in deployment type, the validation error is throw by FSx API model at the client side. It is redundant to do additional validation here. In my opinion, CSI driver code, as a wrapper around the FSx API,  should behave as a passthrough without reinventing any validation logic already implemented in FSx API model.

PR #149, opened 2020-06-28
Revert "Add arm support"

leakingtapan:
Reverts kubernetes-sigs/aws-fsx-csi-driver#147


jqmichael:
Was there a reason ARM support was reverted? @leakingtapan

PR #160, opened 2020-10-12
Add support for HDD Storage Type

gkao123:
Adding support for HDD, a new FSx feature

https://aws.amazon.com/about-aws/whats-new/2020/08/amazon-fsx-lustre-announces-high-performance-hdd-based-shared-storage-compute-workloads/

**What testing is done?** 
End to end testing

### DIFF
--- a/examples/kubernetes/dynamic_provisioning/README.md
+++ b/examples/kubernetes/dynamic_provisioning/README.md
@@ -13,15 +13,19 @@ parameters:
   subnetId: subnet-056da83524edbe641
   securityGroupIds: sg-086f61ea73388fb6b
   deploymentType: PERSISTENT_1
+  storageType: HDD
 ```
 * subnetId - the subnet ID that the FSx for Lustre filesystem should be created inside.
 * securityGroupIds - a common separated list of security group IDs that should be attached to the filesystem
 * deploymentType (Optional) - FSx for Lustre supports three deployment types, SCRATCH_1, SCRATCH_2 and PERSISTENT_1. Default: SCRATCH_1.
 * kmsKeyId (Optional) - for deployment type PERSISTENT_1, customer can specify a KMS key to use.
 * perUnitStorageThroughput (Optional) - for deployment type PERSISTENT_1, customer can specify the storage throughput. Default: "200". Note that customer has to specify as a string here like "200" or "100" etc.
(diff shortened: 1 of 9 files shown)


gkao123:
@leakingtapan Hi Cheng, the tests are failing due to the following error message:
```failed to created filesystem There is already a disk with same ID and different size```.

Could this be related to previous failed test runs?

**Comment by wongma7** [comment]
I think the bug has to do with how we resuse the same fsid forall tests to avoid hitting the 5 minute creation timeout, but we should not need to anymore ever since https://github.com/kubernetes/kubernetes/pull/84140 merged

**Comment by wongma7** [comment]
Never mind, I am not sure the issue, because actually it is the pre-provisioned tests (these tests are SUPPOSED to resuse the same volume) that are failing. will keep digging.

**Comment by wongma7** [comment]
OK, I checked cloudtrail and the error there does not match what we got back?

```
    "errorCode": "IncompatibleParameterError",
    "errorMessage": "The cache type can not be specified for a lustre file system with DeploymentType=SCRATCH_1",
```

Never mind about the volume reuse stuff, since BOTH dynamic provisioning and pre-provisioned tests make a CreateVolume call, and the reuse issue was fixed a long time ago already, that can't be it even though the error message "
      failed to created filesystem There is already a disk with same ID and different size " sounds a lot like it.

File: tests/e2e/conformance_test.go [hunk 1]
```diff
@@ -93,12 +93,16 @@ func (e *fsxDriver) CreateVolume(config *testsuites.PerTestConfig, volType testp
...
 	subnetId := *instance.SubnetId
+	defaultStorageType := "HDD"
+	defaultDriveCacheType := "NONE"
 
 	ctx := context.Background()
 	options := &fsx.FileSystemOptions{
 		CapacityGiB:      3600,
 		SubnetId:         subnetId,
 		SecurityGroupIds: securityGroupIds,
+		StorageType:      defaultStorageType,
```
wongma7, line 104: per my latest comment including the cloudtrail message, also need to set deploymentType to PERSISTENT_1?

gkao123: I included this line of code due to a previous test failure. I realize now that the previous test failure was related to a bug in another code path, so I've gone ahead and removed this parameter in the latest revision.

Issue #151, opened 2020-08-12
FSx driver does not support using different Persistence volumes on the same FSx instance 


aviorma:
/kind bug

**What happened?**
The current FSX driver does not support mounting different persistence volumes on the same FSx instance 
This is blocking us from reusing the same fsx instance on different persistence volumes 

Similar issues with EFS in the past 

https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/100
https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/105
https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/145
https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/167
https://github.com/kubernetes/kubernetes/issues/91556

**What you expected to happen?**

1. A new FSx Persistence volume should creates a dedicated folder on "/" path named with the persistence volume name

2. Each Statefulset pod-pv will be mounted on a new folder on the root tree on the FSx instance** 

**How to reproduce it (as minimally and precisely as possible)?**

1. Install Latest fsx driver on EKS
2. Create different persistence volumes from FSx storage class
3. Running `df -kh` within the container hosting the FSx mount point and see that it has the same mount point for different persistence volumes!!
**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
`Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.6", GitCommit:"d32e40e20d167e103faf894261614c5b45c44198", GitTreeState:"clean", BuildDate:"2020-05-20T13:16:24Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.9-eks-4c6976", GitCommit:"4c6976793196d70bc5cd29d56ce5440c9473648e", GitTreeState:"clean", BuildDate:"2020-07-17T18:46:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}`

- Driver version:

{
  "driverVersion": "v0.4.0-dirty",
  "gitCommit": "4cca48b4b005e50685a9117bec75d76014b8ad72",
  "buildDate": "2020-06-28T06:40:22Z",
  "goVersion": "go1.13.4",
  "compiler": "gc",
  "platform": "linux/amd64"
}

scopej:
[...]

PR #162, opened 2020-10-27
Allow use of an existing Lustre for FSx filesystem

scopej:
This is a new feature, raised in https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/151

This PR is about using an existing Lustre for FSx filesystem across many PersistentVolumes in one namespace.
One part adds in support for mounting a subpath of the filesystem via bind mounts.
The second adds support for setting the FileSystemId in a StorageClass to make setup of PVCs easier.

**What testing is done?** 
I've added relevant unit tests to all sections (And cleaned up the tests some, I'm happy to revert those changes if you'd like, but I think the delta under test is more clear now).

I have run this on a test cluster and applied the example configs and verified it works for my use case.

I HAVE NOT done the e2e tests. I'd like to, but need some help. Is there any documentation for running them? I couldn't get them running from an EC2 instance with basically Administrator access to my AWS account.

**Some questions:**
1. What should the volumeHandle be?
   Currently: shared/<fileSystemId>/<volumeName>
2. Lustre is always mounted rw, because we don't have MountOptions in the StageVolumeRequest, does that affect anything? Does the remote filesystem get mount options? The bind mount will be ro if requested.
3. I don't know much about flock, is there anything I need to do there with the bind mount.
4. Thoughts on how to clean up old volumes? I'm not sure if we can have the controller mount the filesystem to delete. For now, I'm leaving that as an exercise to the user.

### DIFF
--- a/examples/kubernetes/dynamic_provisioning_existing_fs/README.md
+++ b/examples/kubernetes/dynamic_provisioning_existing_fs/README.md
@@ -1,0 +1,54 @@
+## Dynamic Provisioning Example
+This example shows how to use an existing FSx for Lustre filesystem using persistence volume claim (PVC) and consumes it from a pod.
+Each PVC will create a separate folder at the root of the filesystem.
+Update the parameters to match your filesystem.
+
+### Edit [StorageClass](./specs/storageclass.yaml)
+```
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: fsx-sc-0d3a
+provisioner: fsx.csi.aws.com
+parameters:
+  fileSystemId: fs-0d3a9c1160a2ef4ad
+mountOptions:
+  - flock
+```
+
+### Edit [Persistent Volume Claim Spec](./specs/claim.yaml)
+```
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: fsx-claim
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: fsx-sc-03da
+  resources:
+    requests:
+      storage: 6000Gi
+```
+NOTE: `spec.resource.requests.storage` is not used at this time. It must be managed outside of kubernetes.
+
+### Deploy the Application
+Create PVC, storageclass and the pod that consumes the PV:
+```sh
+>> kubectl apply -f examples/kubernetes/dynamic_provisioning_existing_fs/specs/storageclass.yaml
+>> kubectl apply -f examples/kubernetes/dynamic_provisioning_existing_fs/specs/claim.yaml
+>> kubectl apply -f examples/kubernetes/dynamic_provisioning_existing_fs/specs/pod.yaml
+```
+
+### Check the Application uses FSx for Lustre filesystem
+After the objects are created, verify that pod is running:
+
+```sh
+>> kubectl get pods
+```
+
+Also verify that data is written onto FSx for Luster filesystem:
+
+```sh
+>> kubectl exec -ti fsx-app -- tail -f /data/out.txt
+```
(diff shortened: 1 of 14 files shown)


wongma7:
/ok-to-test

re: running e2e tests

You can try the `make test-e2e` rule, that is what the CI executes, however I haven't tried it myself in a while...aside from aws creds which you provided, I am not sure the other requirements, if you run it again could you share logs and I will try to help debug (and update docs). 

You can rely on the PR CI to give you a general signal if stuff broke but it won't help you iterate and add your own test, we'll need to figure out how to run them.

examples/kubernetes/dynamic_provisioning_existing_fs/README.md
```diff
@@ -0,0 +1,54 @@
+## Dynamic Provisioning Example
+This example shows how to use an existing FSx for Lustre filesystem using persistence volume claim (PVC) and consumes it from a pod.
```
wongma7, line 2: s/persistence/persistent

we have this typo in a lot of places !

examples/kubernetes/static_provisioning_with_subpath/README.md
```diff
@@ -0,0 +1,54 @@
...
+  csi:
+    driver: fsx.csi.aws.com
+    volumeHandle: [FileSystemId]
+    volumeAttributes:
+      dnsname: [DNSName]
+      mountname: [MountName]
+      subpath: someDirName
+```
+Replace `volumeHandle` with `FileSystemId`, `dnsname` with `DNSName` and `mountname` with `MountName`. You can get both `FileSystemId`, `DNSName` and `MountName` using AWS CLI:
+When the optional subpath parameter is set, this will create a folder with the name from subpath in the root directory of the FSx filesystem.
```
wongma7, line 29: I am a bit wary about this approach, it blurs the line between static/dynamic provisioning. Strictly speaking I would expect static to mean every property of this volume already exists, including the subpath in it.

What if I set a `subPath` in my pod's volumeMount? If that subPath doesn't exist, the mount fails right?

So the difference is that if i set subpath in volumeAttributes the mount will be guaranteed to work because the driver will create it?

scopej: Yeah, it exists (mostly) in support of dynamic provisioning many pv on one fsx.
Maybe there is some way to avoid it, but it seems that the subPath is on the mount so if I want to dynamically provision separate volumes on the same filesystem, I need a way to curry the intent to the node diver.

wongma7: I got it, I agree this is useful now I am just mulling over the api/syntax.

Your version of `subpath` has two implications, 1 is that the subpath will be created, 2 is that the subpath will be mounted (same as if I put subPath in my volumeMount), I think it would be clearer if these 2 things were split out with clearer intent. 

To solve 2, the EFS driver puts the path in the volumeHandle like fs-id:/path. The issue with using a volumeAttribute to store the path we discovered with EFS is it can lead to Pod launch failure if a Pod tries to consume 2 volumes with the same `volumeHandle` but different attributes. A volumeHandle alone must uniquely identify a volume, so if Kubernetes sees 

volume[0]
volumeHandle: fs-a
volumeAttribute: subpath: a
volume[1]
volumeHandle: fs-a
volumeAttribute: subpath: b

It won't bother to mount volume[1] even though it logically represents a different volume.

But thjis works as users expect:
volume[0]
volumeHandle: fs-a:/a
volume[1]
volumeHandle: fs-a:/b

To solve 1, I would like a volumeAttribute similar to HostPath's Directory or DirectoryOrCreate https://kubernetes.io/docs/concepts/storage/volumes/, and I would like the solution to be the same for all our CSI drivers to be consistent

aviorma:
any estimation for merge and release ?

scopej:
Sorry for the delays, working on CLA approval. 

I think I can move from volumeAttribute to volumeContext to not confuse it with the existing subPath features. Would that work?

PR #173, opened 2021-03-31
Add self to OWNERS

ayberk:
Matthew is a single point of failure, so we need at least one more AWS engineer for approvals.


wongma7:
I don't have power to merge PRs, need to revive CI first https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/179

PR #186, opened 2021-06-07
Add support for Lustre compression

wuxingro:

It adds support for Lustre data compression, a new Lustre feature

https://aws.amazon.com/about-aws/whats-new/2021/05/amazon-fsx-for-lustre-now-supports-data-compression/

**What testing is done?** 

End to end testing

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/kubernetes-sigs/aws-fsx-csi-driver
 
 require (
+	github.com/aws/aws-k8s-tester/e2e/tester v0.0.0-20210607224202-298c17a2707a // indirect
 	github.com/aws/aws-sdk-go v1.38.52
 	github.com/container-storage-interface/spec v1.2.0
 	github.com/golang/mock v1.3.1
@@ -8,6 +9,7 @@ require (
 	github.com/onsi/ginkgo v1.10.1
 	github.com/onsi/gomega v1.7.0
 	google.golang.org/grpc v1.23.1
+	gopkg.in/square/go-jose.v2 v2.2.2 // indirect
 	k8s.io/api v0.17.0
 	k8s.io/apimachinery v0.17.0
 	k8s.io/client-go v0.17.0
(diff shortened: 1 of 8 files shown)


wuxingro:
Hi, I noticed that the e2e test has been failing at the same step (compile e2e) since 3/11.

Job history: https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-aws-fsx-csi-driver-e2e
Error message:
```
Failed to compile e2e:
# github.com/kubernetes-sigs/aws-fsx-csi-driver/tests/e2e
../../../../pkg/mod/github.com/container-storage-interface/spec@v1.2.0/lib/go/csi/csi.pb.go:9:2: missing go.sum entry for module providing package github.com/golang/protobuf/proto (imported by github.com/container-storage-interface/spec/lib/go/csi); to add:
	go get github.com/container-storage-interface/spec/lib/go/csi@v1.2.0
FAIL	github.com/kubernetes-sigs/aws-fsx-csi-driver/tests/e2e [setup failed]
FAIL
```

In my latest diff, I tried to pull from the v1.6.0 branch instead of the master branch of the aws-k8s-tester repo by removing the `@master` annotation in `Makefile`, but I'm still having the same issue. I also took a look at [ebs-csi-driver repo](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/commit/d8c163a070d1089b67df3b5ef39eb0807192ccd0) - seems like they're not pulling from the aws-k8s-tester repo at all - instead they use local `./tests/e2e/` as their test path. I'm wondering if FSx team should do something similar as a workaround.

go.mod
```diff
@@ -1,13 +1,15 @@
 module github.com/kubernetes-sigs/aws-fsx-csi-driver
 
 require (
-	github.com/aws/aws-sdk-go v1.35.7
+	github.com/aws/aws-k8s-tester/e2e/tester v0.0.0-20210607224202-298c17a2707a // indirect
```
wongma7, line 4: we shouldn't need this anymore, can you try running go mod tidy

wongma7:
/lgtm
/approve

just one question about go.mod but CI is passing , so I can fix it in a followup

go.mod
wuxingro: Done. Thanks for catching that!

Issue #191, opened 2021-06-11
Helm chart 1.0


wongma7:
It should be consistent with ebs or efs helm charts  (although ,even ebs and efs are not quite consistent with each other...

wongma7:
https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/190 tolerations

wongma7:
CHANGELOG (WIP):

- Remove controller/nodeService fsxPlugin.image in favor of top-level image value deciding both
- Remove extraArgs, securityContext, podSecurityContext  (may be added back later in different form if needed)
- Bump sidecars (also pull from gcr instead of quay)
- Add imagePullSecrets and dnsPolicy, dnsConfig https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/188

PR #194, opened 2021-06-17
Helm chart 1.0

wongma7:
**Is this a bug fix or adding new feature?** fixes https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/191

**What testing is done?**


wongma7:
• Failure [604.565 seconds]
[fsx-csi-e2e] Dynamic Provisioning with s3 data repository
/home/prow/go/src/github.com/kubernetes-sigs/aws-fsx-csi-driver/tests/e2e/dynamic_provisioning_test.go:111
  should create a volume on demand with s3 as data repository [It]
  /home/prow/go/src/github.com/kubernetes-sigs/aws-fsx-csi-driver/tests/e2e/dynamic_provisioning_test.go:151 

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-fsx-csi-driver/194/pull-aws-fsx-csi-driver-e2e/1416172104553861120

/retest

Makefile
```diff
@@ -66,3 +75,10 @@ image-release:
 .PHONY: push-release
 push-release:
 	docker push $(IMAGE):$(VERSION)
+
+generate-kustomize: bin/helm
+	cd charts/aws-fsx-csi-driver && ../../bin/helm template kustomize . -s templates/csidriver.yaml > ../../deploy/kubernetes/base/csidriver.yaml
```
nckturner, line 80: Is `helm template kustomize` a command?  A cursory search by me didn't find a good reference.

Makefile
```diff
@@ -66,3 +75,10 @@ image-release:
 .PHONY: push-release
 push-release:
 	docker push $(IMAGE):$(VERSION)
+
+generate-kustomize: bin/helm
```
nckturner, line 79: So this generates the kustomize from the helm charts, which are two distinct installation mechanisms, correct?  And after making a change to the helm chart, is it expected for a contributor to run this makefile target?

charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
```diff
@@ -0,0 +1,105 @@
...
+        {{- end }}
+      hostNetwork: true
+      serviceAccountName: {{ .Values.controller.serviceAccount.name }}
+      priorityClassName: system-cluster-critical
+      tolerations:
+          - key: CriticalAddonsOnly
+            operator: Exists
+      containers:
+        - name: fsx-plugin
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```
krmichel, line 38: [This PR](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/988) in EBS is making it so that the version will default to the appVersion from the chart.

charts/aws-fsx-csi-driver/templates/controller-serviceaccount.yaml
```diff
@@ -1,22 +1,23 @@
...
+  name: {{ .Values.controller.serviceAccount.name }}
   labels:
-    {{- include "helm.labels" . | nindent 4 }}
-  {{- with .Values.serviceAccount.annotations }}
+    {{- include "aws-fsx-csi-driver.labels" . | nindent 4 }}
+  {{- with .Values.controller.serviceAccount.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+{{- end }}
```
krmichel, line 12: Moving this end here will create the ClusterRole and ClusterRoleBinding regardless of whether the service account is created.  I am not 100% sure what will happen with the ClusterRoleBinding since the service account it is referencing won't exist if `.Values.controller.serviceAccount.create` is false.  Seems like this end should move back to the end of the template.

charts/aws-fsx-csi-driver/templates/node-daemonset.yaml
```diff
@@ -1,32 +1,38 @@
...
-  name: {{ include "helm.fullname" . }}-daemonset
+  name: fsx-csi-node
   labels:
-    {{- include "helm.labels" . | nindent 4 }}
+    {{- include "aws-fsx-csi-driver.labels" . | nindent 4 }}
 spec:
   selector:
     matchLabels:
-      {{- include "helm.selectorLabels" . | nindent 6 }}-daemonset
+      app: fsx-csi-node
```
krmichel, line 10: Adding this will cause upgrade issues since matchLabels is immutable. It looks like maybe it is being added to ensure that is continues to be included on the kustomize manifests. If that is the case, I would move if from here to an else in the selectorLabels helper template.

charts/aws-fsx-csi-driver/templates/node-daemonset.yaml
```diff
@@ -1,32 +1,38 @@
 apiVersion: apps/v1
```
krmichel, line 1: Since this is a daemonset and meant to run on every node it should probably have some default tolerations with the ability to set more.  EBS handles it like this:
```
      tolerations:
        {{- if .Values.node.tolerateAllTaints }}
        - operator: Exists
        {{- else }}
        - key: CriticalAddonsOnly
          operator: Exists
        - operator: Exists
          effect: NoExecute
          tolerationSeconds: 300
        {{- end }}
        {{- with .Values.node.tolerations }}
        {{- toYaml . | nindent 8 }}
        {{- end }}
```

charts/aws-fsx-csi-driver/templates/node-serviceaccount.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: {{ .Values.node.serviceAccount.name }}
+  labels:
+    {{- include "aws-fsx-csi-driver.labels" . | nindent 4 }}
+  {{- with .Values.node.serviceAccount.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+{{- end }}
+---
```
krmichel, line 13: Does the service account for the node not need any RBAC permissions?

deploy/kubernetes/base/controller-deployment.yaml
```diff
@@ -1,34 +1,37 @@
...
   name: fsx-csi-controller
-  namespace: kube-system
+  labels:
+    app.kubernetes.io/name: aws-fsx-csi-driver
 spec:
   replicas: 2
   selector:
     matchLabels:
       app: fsx-csi-controller
+      app.kubernetes.io/name: aws-fsx-csi-driver
```
krmichel, line 14: I don't know if this is a concern but this will need intervention on a new apply since matchLabels are immutable.

charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
```diff
@@ -0,0 +1,105 @@
...
+      nodeSelector:
+        kubernetes.io/os: linux
+        {{- with .Values.controller.nodeSelector }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      hostNetwork: true
+      serviceAccountName: {{ .Values.controller.serviceAccount.name }}
+      priorityClassName: system-cluster-critical
+      tolerations:
+          - key: CriticalAddonsOnly
```
krmichel, line 34: These are indented more than normal and it causes a change in the kustomize manifest

deploy/kubernetes/base/controller-deployment.yaml
```diff
@@ -47,21 +50,37 @@ spec:
...
+            periodSeconds: 2
+            failureThreshold: 5
         - name: csi-provisioner
-          image: quay.io/k8scsi/csi-provisioner:v1.3.0
+          image: public.ecr.aws/eks-distro/kubernetes-csi/external-provisioner:v2.1.1-eks-1-18-2
           args:
-            - --timeout=5m
             - --csi-address=$(ADDRESS)
-            - --v=5
-            - --enable-leader-election
```
krmichel, line 56: I see that the image is changing, just wanted to make sure dropping the leader election flags is intended.

deploy/kubernetes/base/node-daemonset.yaml
```diff
@@ -1,32 +1,36 @@
...
 metadata:
   name: fsx-csi-node
-  namespace: kube-system
+  labels:
+    app.kubernetes.io/name: aws-fsx-csi-driver
 spec:
   selector:
     matchLabels:
       app: fsx-csi-node
+      app.kubernetes.io/name: aws-fsx-csi-driver
```
krmichel, line 13: I don't know if this is a concern but this will need intervention on a new apply since matchLabels are immutable.

krmichel (changes requested):

charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
```diff
@@ -0,0 +1,105 @@
...
+kind: Deployment
+metadata:
+  name: fsx-csi-controller
+  labels:
+    {{- include "aws-fsx-csi-driver.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.controller.replicaCount }}
+  selector:
+    matchLabels:
+      app: fsx-csi-controller
```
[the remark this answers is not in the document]
wongma7: I'll test out the upgrade case, since I am bumping major version of helm chart I am not sure without testing how helm deals with it. I just cargo culted how ebs does it TBH: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/controller.yaml#L12

charts/aws-fsx-csi-driver/templates/controller-serviceaccount.yaml
wongma7: It's intentional to match how ebs does it https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/serviceaccount-csi-controller.yaml (ebs splits up the serviceaccount from the associated RBAC bindings whereas here it's same fille...I kind of prefer keeping serviceaccount and associated RBAC in same file, but for ebs there are a LOT of RBAC things). The reasoning is that even if .Values.controller.serviceAccount.create is false, .Values.controller.serviceAccount.name is respected. An example use-case is on EKS if I take advantage of the "IAM for service accounts" feature then I will probably want to create the serviceAccount myself because I need it to have a special annotation. But I still need the ClusterRole and ClusterRoleBinding to exist for my serviceAccount whether it is created by me or by helm

charts/aws-fsx-csi-driver/templates/node-serviceaccount.yaml
wongma7: no, but it will in the future if we implement instance metadata workaround where the node pod gets the instance id and such by `Getting` its `Node` versus trying to query instance metadata on local node.

deploy/kubernetes/base/controller-deployment.yaml
wongma7: no, I added back

wongma7:
I resolved comments I think have been addressed by latest 5 commits, the one open question is how to deal with matchLabels which I will test out.  appreciate the thorough review

charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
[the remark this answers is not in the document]
wongma7: this change is safe because helm somehow knows to delete and recreate the deployment which also deletes and recreates the pods..

1. I installed the current helm chart v0.1.0 using the README command 
  a. (I manually edited the deployment and daemonset to have hostNetwork and to point to image 0.4.0 instead of 0.3.0 because they were crashlooping as I have instance metadata disabled)
2. I ran `helm upgrade --install aws-fsx-csi-driver  ./charts/aws-fsx-csi-driver`
3. helm recreated the deployment
```
k get deployment
NAME                 READY   UP-TO-DATE   AVAILABLE   AGE
fsx-csi-controller   1/1     1            1           57s
```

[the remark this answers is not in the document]
krmichel: I didn't notice that the name of the deployment was changing.  That explains the delete and create behavior.

charts/aws-fsx-csi-driver/templates/controller-serviceaccount.yaml
krmichel: Personally, I would just add the annotation to `controller.serviceAccount.annotations`, which is what I did for the EBS chart.  I do, however, see what you are saying about some users potentially wanting to use a pre-existing service account.

If this chart isn't creating a service account and therefore expects that one already exists it might be worth using the lookup function to validate that it does exist and fail if it doesn't.
https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function

deploy/kubernetes/base/controller-deployment.yaml
krmichel: This may still be the case since the deployment name isn't changing.  Did you test the upgrade with kustomize?

krmichel (changes requested):
Let me know what you think about trying to validate that the service account exists.  Did you test the kustomize based upgrade?  It looks like the names of those resources are changing.  I haven't used kustomize but based on past kubectl behavior it seems like the match labels could be an issue.

wongma7:
latest commit gives lookup function a try. it seems to work but the error message is quite ugly, I'm not sure if I called `fail` correctly?

```
$ helm upgrade --install aws-fsx-csi-driver  ./charts/aws-fsx-csi-driver --set controller.serviceAccount.create=false
Release "aws-fsx-csi-driver" does not exist. Installing it now.
Error: template: aws-fsx-csi-driver/templates/controller-serviceaccount.yaml:15:8: executing "aws-fsx-csi-driver/templates/controller-serviceaccount.yaml" at <fail (printf "create serviceaccount %s/%s or set .controller.serviceaccount.create true" .Release.Namespace .Values.controller.serviceAccount.name)>: error calling fail: create serviceaccount default/fsx-csi-controller-sa or set .controller.serviceaccount.create true
~/g/s/g/k/aws-fsx-csi-driver (helmchart1.0|✚1) [1] $ k create serviceaccount fsx-csi-controller-sa
serviceaccount/fsx-csi-controller-sa created
~/g/s/g/k/aws-fsx-csi-driver (helmchart1.0|✚1) $ helm upgrade --install aws-fsx-csi-driver  ./charts/aws-fsx-csi-driver --set controller.serviceAccount.create=false
Release "aws-fsx-csi-driver" does not exist. Installing it now.
NAME: aws-fsx-csi-driver
LAST DEPLOYED: Wed Aug  4 12:59:31 2021
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

wongma7:
the kustomize upgrade indeed failed to apply. however it worked with --force, I suppose I can add this to the readme installation instructions as upgrade note, same as for helm (working on that now)

krmichel:
> latest commit gives lookup function a try. it seems to work but the error message is quite ugly, I'm not sure if I called `fail` correctly?
> 
> ```
> $ helm upgrade --install aws-fsx-csi-driver  ./charts/aws-fsx-csi-driver --set controller.serviceAccount.create=false
> Release "aws-fsx-csi-driver" does not exist. Installing it now.
> Error: template: aws-fsx-csi-driver/templates/controller-serviceaccount.yaml:15:8: executing "aws-fsx-csi-driver/templates/controller-serviceaccount.yaml" at <fail (printf "create serviceaccount %s/%s or set .controller.serviceaccount.create true" .Release.Namespace .Values.controller.serviceAccount.name)>: error calling fail: create serviceaccount default/fsx-csi-controller-sa or set .controller.serviceaccount.create true
> ~/g/s/g/k/aws-fsx-csi-driver (helmchart1.0|✚1) [1] $ k create serviceaccount fsx-csi-controller-sa
> serviceaccount/fsx-csi-controller-sa created
> ~/g/s/g/k/aws-fsx-csi-driver (helmchart1.0|✚1) $ helm upgrade --install aws-fsx-csi-driver  ./charts/aws-fsx-csi-driver --set controller.serviceAccount.create=false
> Release "aws-fsx-csi-driver" does not exist. Installing it now.
> NAME: aws-fsx-csi-driver
> LAST DEPLOYED: Wed Aug  4 12:59:31 2021
> NAMESPACE: default
> STATUS: deployed
> REVISION: 1
> TEST SUITE: None
> ```

Looks like a correct use of fail.

krmichel (approved):

PR #198, opened 2021-08-06
Release 0.5.0 part 1/3: CHANGELOG, README, Makefile

wongma7:

 part 1 is to update docs and Makefile such that we can tag this commit and trigger image builds.

parts 2 and 3 will follow the image builds.

(I am following https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/RELEASE.md)

**What testing is done?**


jmwurst:
LGTM!

PR #203, opened 2021-08-14
make the helm chart secret name configurable

cdemers:

It's a feature, a small enhancement to the Helm chart.

It's backward compatible, and it makes the K8s secret name (previously hardcoded to `aws-secret`) configurable.  It's pretty useful to be able to name your secrets something other than _aws-secret_ to clarify what the secret is for when looking at it from the K8s side, and to be able to figure out which secret is the one with the IAM keys for the aws-fsx-csi-driver when you have many secrets in the same namespace.

**What testing is done?** 

Running the helm template command returns exactly the same result before and after (unless you override the Helm value, then the value does changes in the deployment section to whatever you specified).

**Also**

Maybe the same kind of configurability might be done on the _kustomize_ side of things, but I prefer to make this a separate PR.

### DIFF
--- a/charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
@@ -51,13 +51,13 @@ spec:
             - name: AWS_ACCESS_KEY_ID
               valueFrom:
                 secretKeyRef:
-                  name: aws-secret
+                  name: {{ .Values.controller.secretName }}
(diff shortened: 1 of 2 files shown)


jmwurst:
lgtm!

cdemers:
Is there anything I could do to help this PR move forward? I'm open to any suggestions

jmwurst:
@wongma7 I believe this is fine to approve.

PR #204, opened 2021-08-25
Add multiple StorageClass resources from values

ArieLevs:
feature

- [x] this feature will provide the ability to deploy multiple `StorageClass` resources part of helm install process, this is needed when bootstrapping  a cluster for example, so an automated process will be able to install the chart + storage class in a single command.
- [x] provide advanced affinity configuration for controller/nodes, for example controller pods should be distributed to non application nodes, and daemonset should be distributed to nodes that also run pods that need FSX

**What testing is done?** 
`helm template`
`helm lint`
and installation attempt on local k8s cluster


ArieLevs:
@leakingtapan @wongma7

ArieLevs:
@wongma7

PR #222, opened 2021-11-04
Refactoring Kubernetes labels and add custom labels

nlamirault:

Feature

Support to add custom labels into metadata of all resources

**What testing is done?**

### DIFF
--- a/charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
@@ -8,12 +8,10 @@ spec:
   replicas: {{ .Values.controller.replicaCount }}
   selector:
     matchLabels:
-      app: fsx-csi-controller
       {{- include "aws-fsx-csi-driver.selectorLabels" . | nindent 6 }}
   template:
     metadata:
       labels:
-        app: fsx-csi-controller
         {{- include "aws-fsx-csi-driver.labels" . | nindent 8 }}
     spec:
       {{- if .Values.imagePullSecrets }}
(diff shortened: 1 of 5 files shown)


charts/aws-fsx-csi-driver/values.yaml
```diff
@@ -7,6 +7,10 @@ image:
   tag: v0.6.0
   pullPolicy: IfNotPresent
 
+# -- Custom labels to add into metadata
+customLabels: {}
```
jmwurst, line 11: The problem with this implementation is that it does not allow for distinct labels for the node/controller pods, which is important because the Controller and Node services perform different functions.

nlamirault: Like that ? we could customize all resources, node and controller : 
```yaml
customLabels:
  global: {}
  node: {}
  controller: {}
```

wongma7: we can always add `controller.customLabels` and `node.customLabels` later. so in sum

`customLabels`
`controller.customLabels` (if needed)
`node.customLabels` (if needed)

charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
```diff
@@ -8,12 +8,10 @@ spec:
   replicas: {{ .Values.controller.replicaCount }}
   selector:
     matchLabels:
-      app: fsx-csi-controller
```
wongma7, line 11: we should keep this, it's maybe redundant with the new `part-of` label but nonzero users have probably come to depend on it existing (personally i find it useful for kubectl get)

also EBS still has it so my top priority is staying consistent to avoid more divergence! https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/controller.yaml#L16
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/974

PR #234, opened 2021-12-30
Release 0.8.0 part 1/3: CHANGELOG, README, Makefile

khoang98:
Bug fix

Update AL2 base image again & create release with updated helm

**What testing is done?** 
None

### DIFF
--- a/CHANGELOG-0.x.md
+++ b/CHANGELOG-0.x.md
@@ -1,3 +1,11 @@
+# v0.8.0
+
+### Misc.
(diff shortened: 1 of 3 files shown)


CHANGELOG-0.x.md
```diff
@@ -1,3 +1,11 @@
+# v0.8.0
+
+### Misc.
+* Release 0.7.0 part 3/3 ([#224](https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/224), [@khoang98](https://github.com/khoang98))
```
wongma7, line 4: I would remove these but at the same time I don't mind keeping the list exhaustive and just keeping this stuff under 'Misc.' either

khoang98: Ah ok makes sense. Will leave for now but if it gets too unreadable, can remove these in future

PR #238, opened 2022-01-28
Add fsx:TagResource to node IAM policy

wongma7:
**Is this a bug fix or adding new feature?** /bug
fix

**What is this PR about? / Why do we need it?** Same as https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/215 . Not sure how / when this became a requirement or how we have avoided it until now but it seems necessary based on logs of latest failures https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-fsx-csi-driver/237/pull-aws-fsx-csi-driver-e2e/1486791696619409408

**What testing is done?** 
PR CI test

### DIFF
--- a/hack/kops-patch.yaml
+++ b/hack/kops-patch.yaml
@@ -17,7 +17,8 @@ spec:
             "s3:ListBucket",
             "fsx:CreateFileSystem",
             "fsx:DeleteFileSystem",
-            "fsx:DescribeFileSystems"
+            "fsx:DescribeFileSystems",
+            "fsx:TagResource"
           ],
           "Resource": ["*"]
         }


hack/kops-patch.yaml
```diff
@@ -17,7 +17,8 @@ spec:
             "s3:ListBucket",
             "fsx:CreateFileSystem",
             "fsx:DeleteFileSystem",
-            "fsx:DescribeFileSystems"
+            "fsx:DescribeFileSystems",
+            "fsx:TagResource"
```
nckturner, line 21: Update readme to add this permission?

nckturner (approved):

PR #246, opened 2022-04-08
Add Idempotent check for mounting node volume

amankhunt:
Bug fix
Basically this fix will enable CSI driver to check if target mount is already mounted. client will not face error such as  "x is already mounted"
**What testing is done?** 
manually end to end tested

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -100,10 +100,24 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 		return nil, status.Errorf(codes.Internal, "Could not create dir %q: %v", target, err)
 	}
 
-	klog.V(5).Infof("NodePublishVolume: mounting %s at %s with options %v", source, target, mountOptions)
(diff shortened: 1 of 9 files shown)


pkg/driver/node.go
```diff
@@ -100,10 +100,24 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 		return nil, status.Errorf(codes.Internal, "Could not create dir %q: %v", target, err)
 	}
 
-	klog.V(5).Infof("NodePublishVolume: mounting %s at %s with options %v", source, target, mountOptions)
-	if err := d.mounter.Mount(source, target, "lustre", mountOptions); err != nil {
-		os.Remove(target)
-		return nil, status.Errorf(codes.Internal, "Could not mount %q at %q: %v", source, target, err)
+	//klog.V(5).Infof("NodePublishVolume: mounting %s at %s with options %v", source, target, mountOptions)
```
khoang98, line 103: Can just delete this instead of commenting out and leaving it here

amankhunt: done, thanks for pointing out

PR #254, opened 2022-08-23
Use k8s.io/mount-utils instead of k8s.io/utils

khoang98:
Bug fix

We wanted to fix an issue where unmounting a mount point that was already unmounted would throw an error. Refactoring the driver to use mount-utils pulls in this check for us.

**What testing is done?** 
Ran make test locally, build and locally deployed driver (create pod, manually unmount, delete pod).


wongma7:
/ok-to-test
/lgtm
/approve
(kinda surprised it is so straightforward i thought their APIs diverged or something but i guess not )

PR #262, opened 2022-10-02
Added templating for CSIDriver object configuration

Mberga14:

New feature: Adding templating for configuring the CSIDriver object.

To run pods with non-root users and to avoid usage of initContainers that chown the volume, we want to be able to let k8s traverse the filesystem and change the ownership for us. This currently cannot be done as the CSIDriver object always sets defaults, specifically `fsGroupPolicy`

**What testing is done?** 

Tested by applying the new chart version to a cluster running k8s v 1.23.12. Was able to change `fsGroupPolicy` to `File`, allowing non-root pods to have read/write access to the provisioned volume

### DIFF
--- a/charts/aws-fsx-csi-driver/templates/csidriver.yaml
+++ b/charts/aws-fsx-csi-driver/templates/csidriver.yaml
@@ -3,12 +3,5 @@ kind: CSIDriver
 metadata:
   name: fsx.csi.aws.com
 spec:
-  attachRequired: {{ .Values.csidriver.attachRequired }}
+  attachRequired: false
   fsGroupPolicy: {{ .Values.csidriver.fsGroupPolicy }}
-  podInfoOnMount: {{ .Values.csidriver.podInfoOnMount }}
-  requiresRepublish: {{ .Values.csidriver.requiresRepublish }}
-  storageCapacity: {{ .Values.csidriver.storageCapacity }}
-  volumeLifecycleModes: 
-  {{- range .Values.csidriver.volumeLifecycleModes }}
-  - {{.}}
-  {{- end }}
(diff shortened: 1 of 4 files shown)


charts/aws-fsx-csi-driver/templates/csidriver.yaml
```diff
@@ -3,4 +3,12 @@ kind: CSIDriver
 metadata:
   name: fsx.csi.aws.com
 spec:
-  attachRequired: false
+  attachRequired: {{ .Values.csidriver.attachRequired }}
```
olemarkus, line 6: Most of these settings only define properties of the CSI driver. I don't think it makes sense to have these configurable

charts/aws-fsx-csi-driver/templates/csidriver.yaml
```diff
@@ -3,4 +3,12 @@ kind: CSIDriver
...
+  fsGroupPolicy: {{ .Values.csidriver.fsGroupPolicy }}
```
olemarkus, line 7: This one is an exception, I think. `ReadWriteOnceWithFSType` effectively means that kubelet won't change ownership of the files. Changing this to `File` will cause kubelet to change owner every time this is mounted, and if mounted on many nodes at the same time can lead to interesting situations such as deadlocks.

Its however fairly safe _if_ every mounting Pod has the same `securityContext` _and_ it includes `fsGroupChangePolicy: "OnRootMismatch"`

We may want to document this somewhere.

olemarkus (changes requested):
I suggest making only `fsGroupPolicy`.

charts/aws-fsx-csi-driver/templates/csidriver.yaml
Mberga14: I agree. For now, I put it in the `changelog.md` file, but this is not ok. Maybe it would make sense to create a `README.md` for the helm chart? 

Also: removed the configuration items except `fsGroupPolicy`

olemarkus:
/ok-to-test

Looks good to me. Let's just wait a few more days for the other maintainers.

Issue #247, opened 2022-04-11
Add a subdirectory provisioner mechanism


colearendt:
**Is your feature request related to a problem?/Why is this needed**

FSx Lustre filesystems are large (minimum 1.2T). As a result, it would be nice to be able to couple related applications within a particular FSx Lustre filesystem. 

/feature

**Describe the solution you'd like in detail**

For instance:

- Create a StorageClass that targets an existing / static FSx Lustre filesystem
- When a PVC is requested against that storage class, a directory is created within the root of the system
- The PVC is filled / bound to this directory and made available to the application
- As a result, several applications have mutually exclusive access to a subdirectory of the FSx Lustre filesystem

**Describe alternatives you've considered**

- Independent FSx Lustre filesystem per application (cost prohibitive)
- Use several PVs (one per application) to target the same filesystem at different `mountname`s (mount sub-paths)
    - Select the corresponding PV with the application's PVC
    - Less than ideal because requires a PV per application
    - Also requires the subdirectories to be created manually within the FSx Lustre filesystem (or the mount / PVC binding will not succeed)

**Additional context**

This feature is prompted by the [`nfs-subdir-external-provisioner`'s behavior](https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner) (although nfs-server-provisioner works much the same way too).

pat-s:
@jacobwolfaws Any news why it has been moved to 'frozen'?

PR #272, opened 2022-10-26
Subdirectory provisioning

KanchiShimono:

New feature.
Attempt to implement subdirectory provisioning requested in #247.

PR summary:

- Add new dynamic provisioning mode (subdirectory provisioning) that creating subdirectory under existing FSx file system.
- Add new voulme ID form for subdirectory provisiong to be able to delete subdirectory in DeleteVolume method.
- Add new parameters (dnsname, mountname and basedir) of CreateVolumeRequest.
  - If dnsname was set, the volume is provisioned as subdirectory mode.
  - Subdirectory is created under the basedir.
  - basedir is mounted into controller plugin container to create subdirectory.
  - subdirectory is mounted into volume claimed container.
- Add new volume context for NodePublishVolume/NodeUnPublishVolume.

Why do we need?:

- It's useful for usecase that required hundreds GB storage from multiple pod. (Can share form multiple pod)
- Fast provisiong. (No need to wait for creating new file system.)
- Enable to mount subdirectory on static provisioning.

**What testing is done?** 

Run unit test on my local machine by running `make test`.
Also deploy and check on my EKS cluster.


charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
```diff
@@ -41,6 +41,8 @@ spec:
         {{- end }}
       containers:
         - name: fsx-plugin
+          securityContext:
+            privileged: true
```
jacobwolfaws, line 45: How does this change affect subdirectory provisioning and why is it needed?

deploy/kubernetes/base/controller-deployment.yaml
```diff
@@ -31,6 +31,8 @@ spec:
           tolerationSeconds: 300
       containers:
         - name: fsx-plugin
+          securityContext:
+            privileged: true
```
jacobwolfaws, line 35: same comment as here: https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/272/files#r1059149727

deploy/kubernetes/base/csidriver.yaml
```diff
@@ -6,3 +6,4 @@ metadata:
   name: fsx.csi.aws.com
 spec:
   attachRequired: false
+  fsGroupPolicy: ReadWriteOnceWithFSType
```
jacobwolfaws, line 9: Based on the documentation: https://kubernetes-csi.github.io/docs/support-fsgroup.html it seems like not specifying the fsGroupPolicy policy will default to ReadWriteOnceWithFSType. Is this change needed & could this have any impact on any usecases for csi driver?

kibblerz:
In theory, couldn't a subdirectory be mounted by simply mounting the PV, with a mountname like "mount/subfolder"? I don't have FSx running to test this myself sadly

jacobwolfaws:
> In theory, couldn't a subdirectory be mounted by simply mounting the PV, with a mountname like "mount/subfolder"? I don't have FSx running to test this myself

Hi @kibblerz, from my understanding you can mount a subdirectory like you suggested, but the idea behind this PR is being able to provision the subdirectory using the csi driver.

pkg/driver/subdir_provisioner.go
```diff
@@ -0,0 +1,173 @@
...
+			if err := os.RemoveAll(target); err != nil {
+				klog.Warningf("Could not delete %q: %v", target, err)
+			}
+		} else {
+			klog.Warningf("Could not unmount %q: %v", target, err)
+		}
+	}()
+
+	// Delete subDir (dnsname@tcp:/mountname/baseDir/subDir) under the tempMountPathPrefix/tempUUID
+	if err := os.RemoveAll(filepath.Join(target, fsxVolume.subDir)); err != nil {
```
wongma7, line 168: subDir and baseDir may need validation for relative paths. what happens if I set subDir to nonsense like ../

examples/kubernetes/static_provisioning/README.md
```diff
@@ -18,26 +20,54 @@ spec:
...
     driver: fsx.csi.aws.com
+    # mount root of filesystem
     volumeHandle: [FileSystemId]
+    # mount subdir of filesystem
+    # volumeHandle: [DNSName:MountName:basedir:subdir:uuid]
     volumeAttributes:
-      dnsname: [DNSName] 
+      dnsname: [DNSName]
       mountname: [MountName]
+      # set the basedir and subdir if you wanted to mount subdir
```
wongma7, line 30: in kubernetes the volumehandle must uniquely identify the volume. if you try to mount two PVC+PV on the same node and they have the same volumehandle but different volumeattributes.subdir, I don't think it works. this is why in EFS we removed the path volumeattribute and stuffed everything into the volumehandle like [FileSystemId]:[Path]  instead. https://github.com/kubernetes/kubernetes/blob/497e0dfd33b180fcc3855aad794d87f0e429c82c/pkg/volume/csi/csi_plugin.go#L316

PR #276, opened 2022-11-17
upgrade kubernetes dependencies to v0.22.3

jacobwolfaws:
bug fix
updating kubernetes dependencies, corresponding packages and pack calls to move to v0.22.3. We're currently using a deprecated kubernetes version that contains known CVEs
**What testing is done?** 
manual


olemarkus:
Any reason you stopped at 0.22.3?

jacobwolfaws:
> Any reason you stopped at 0.22.3?

I chose 0.22.3 to keep us in line with the aws-efs-csi-driver + address known CVEs with 0.22.2 and below. I'm not opposed to updating the dependencies further, but that change would have a separate goal than this one.

olemarkus:
Fair enough. I was looking to go for something higher, at least a supported k8s version, myself. I know 1.25 has some challenges due to ginkgo2 requirements, but at least 1.24 should be fine.

/lgtm
/approve

PR #277, opened 2022-11-18
Update import path to match the go module name

olemarkus:
/kind bug

When the module name changed, this script was not updated so mock generation fails.

**What testing is done?** 
Not really applicable


jacobwolfaws:
Is this something we want to be included in the new csi driver release?

PR #278, opened 2022-11-18
Update workflow to push image to ECR public instead of DockerHub

siddharthsalot:
New feature

The workflow will use an Open ID connect IAM identity provider to publish images to ECR public repository: https://gallery.ecr.aws/fsx-csi-driver/aws-fsx-csi-driver Images hosted in ECR public will allow customers to download images without a substantial throttling. This requirement is led by a combination of ECR and security teams to migrate all AWS owned images to ECR public.

**What testing is done?** 
The updated workflow has been tested on a forked branch in a dev AWS account.
Workflow execution: https://github.com/siddharthsalot/aws-fsx-csi-driver/actions/runs/3299837518/jobs/5443751149
Dev ECR public library: https://gallery.ecr.aws/g5x7u5l2/aws-fsx-csi-driver

Prod ECR public library: https://gallery.ecr.aws/fsx-csi-driver/aws-fsx-csi-driver

### DIFF
--- a/charts/aws-fsx-csi-driver/CHANGELOG.md
+++ b/charts/aws-fsx-csi-driver/CHANGELOG.md
@@ -1,4 +1,8 @@
 # Helm chart
+
+# v1.5.0
+* Use driver 0.9.0
+
 # v1.4.4
 * Use driver 0.8.3
 
--- a/charts/aws-fsx-csi-driver/Chart.yaml
+++ b/charts/aws-fsx-csi-driver/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: "0.8.3"
+appVersion: "0.9.0"
 name: aws-fsx-csi-driver
 description: A Helm chart for AWS FSx for Lustre CSI Driver
-version: 1.4.4
+version: 1.5.0
 kubeVersion: ">=1.17.0-0"
 home: https://github.com/kubernetes-sigs/aws-fsx-csi-driver
 sources:
--- a/charts/aws-fsx-csi-driver/values.yaml
+++ b/charts/aws-fsx-csi-driver/values.yaml
@@ -4,7 +4,7 @@
 
 image:
   repository: amazon/aws-fsx-csi-driver
-  tag: v0.8.3
+  tag: v0.9.0
   pullPolicy: IfNotPresent
 
 csidriver:
--- a/deploy/kubernetes/base/controller-deployment.yaml
+++ b/deploy/kubernetes/base/controller-deployment.yaml
@@ -31,7 +31,7 @@ spec:
           tolerationSeconds: 300
       containers:
         - name: fsx-plugin
-          image: amazon/aws-fsx-csi-driver:v0.8.3
+          image: amazon/aws-fsx-csi-driver:v0.9.0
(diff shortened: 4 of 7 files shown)


olemarkus:
Can we please add `registry.k8s.io` promotion before looking at things like ECS?

siddharthsalot:
Hey Markus, thank you for taking a look.

We are trying to move away from DockerHub due to availability and security concerns. We have created our own verified ECR public repository and have synced all existing docker images from dockerhub to ECR public.

Let me know if there are any high-level concerns with this approach.

deploy/kubernetes/overlays/stable/ecr-public/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+bases:
+  - ../../base
+images:
+  - name: 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/aws-fsx-csi-driver
```
nckturner, line 6: What does this do? (why the reference to `602401143452` here)

nckturner (commented):
Also generally you want to bump the chart version: https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/master/charts/aws-fsx-csi-driver/Chart.yaml#L5

nckturner:
/ok-to-test

Are there any tests for the helm chart?

nckturner:
> Can we please add registry.k8s.io promotion before looking at things like ECR?

I think we probably want both.  From AWS perspective, moving from dockerhub to ECR public is important, from the communities perspective it would be great to have artifacts and images published to the CNCF registry.   Can't comment on priority but it doesn't make sense to me to block moving to ECR on the CNCF registry stuff, nor would it make sense to block CNCF registry on ECR, unless there is some dependency I'm unaware of.

nckturner:
@siddharthsalot can you sign the CLA and squash your commits?

olemarkus:
> Hey Markus, thank you for taking a look.
> 
> We are trying to move away from DockerHub due to availability and security concerns. We have created our own verified ECR public repository and have synced all existing docker images from dockerhub to ECR public.
> 
> Let me know if there are any high-level concerns with this approach.

I have nothing against moving away from dockerhub. But I think the promotion process K8s infra has is pretty solid and better than just publishing directly from github tags.

olemarkus:
> > Can we please add registry.k8s.io promotion before looking at things like ECR?
> 
> I think we probably want both. From AWS perspective, moving from dockerhub to ECR public is important, from the communities perspective it would be great to have artifacts and images published to the CNCF registry. Can't comment on priority but it doesn't make sense to me to block moving to ECR on the CNCF registry stuff, nor would it make sense to block CNCF registry on ECR, unless there is some dependency I'm unaware of.

They don't block each other at all. EBS driver already does both, I believe.

deploy/kubernetes/overlays/stable/ecr-public/kustomization.yaml
jmwurst: That value is the account ID owning the repos with Amazon EKS-vended container images - https://docs.aws.amazon.com/eks/latest/userguide/add-ons-images.html

PR #283, opened 2022-12-06
Bump sidecar versions

markapruett:
Bumps sidecars to a more recent version

I'm bumping the sidecar images to version 1.23 to get the conversation started on updating the images used. 1.18 has been out of support for some time. I noticed 1.18 was almost out of support when it was updated last so you may not want to go all the way to 1.23. 

We could also go up to 1.24, but the best would be to use the correct version based on the k8s version being used. Adding  switching logic based on version in helm is easy enough, but first we'd need to rethink the approach of building a single version for all clusters, and move to building different versions for each version of K8s supported by EKS. Short of that, this PR addresses several security and support issues from running old versions. 

**What testing is done?** 
No testing yet


jacobwolfaws (approved):
/approve

PR #285, opened 2022-12-08
Update container-image workflow to publish image to ECR public

siddharthsalot:
New Feature

As per a recent AWS wide campaign, we want to move away from dockerhub and use AWS ECR public repository to host aws-fsx-csi-driver docker images. ECR public provides many advantages over dockerhub wherein customers can request images at higher rates without getting throttled. Amazon ECR will also provide additional security for the images and would provide a better user experience for the customer.

**What testing is done?** 
The testing was conducted on a forked branch. Workflow executed on forked branch: https://github.com/siddharthsalot/aws-fsx-csi-driver/actions/runs/3498761218/jobs/5859534269
ECR public gallery created from Dev account: https://gallery.ecr.aws/g5x7u5l2/aws-fsx-csi-driver

**Note**:
* Originally this change was published as part of PR: https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/278. The original PR was discarded and I created a new one.
* The chart version and the image version need not be updated because this is just a workflow change and does not introduce any new changes to the docker image itself.
* No helm-chart tests were executed as part of testing.

### DIFF
--- a/.github/workflows/container-image.yaml
+++ b/.github/workflows/container-image.yaml
@@ -1,11 +1,15 @@
 name: Container Images
 
 on: push
+env:
+  REGION : "us-east-1"
 jobs:
   build:
     # this is to prevent the job to run at forked projects
     if: github.repository == 'kubernetes-sigs/aws-fsx-csi-driver'
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


jacobwolfaws:
So if I'm understanding this correctly, we'd be moving away entirely from publishing to dockerhub and would instead be publishing it directly to the public ECR repository. Is there a reason why we can't/shouldn't do both?

deploy/kubernetes/overlays/stable/ecr-public/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+bases:
+  - ../../bases
+images:
+  - name: 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/aws-fsx-csi-driver
```
wongma7, line 6: i don't think this works, because the bases https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/master/deploy/kubernetes/base/node-daemonset.yaml#L36 and https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/master/deploy/kubernetes/base/controller-deployment.yaml#L34 say amazon/aws-fsx-csi-driver still. It should say `name: amazon/aws-fsx-csi-driver` I think in order to find/replace amazon/aws-fsx-csi-driver with public.ecr.aws/fsx-csi-driver/aws-fsx-csi-driver

.github/workflows/container-image.yaml
```diff
@@ -14,24 +18,28 @@ jobs:
...
+      - name: Login to Amazon ECR Public
+        id: login-ecr-public
+        uses: aws-actions/amazon-ecr-login@v1
+        with:
+          registry-type: public
+      - name: Build, tag, and push docker image to Amazon ECR Public Repository
+        env:
+          REGISTRY: ${{ steps.login-ecr-public.outputs.registry }}
+          REGISTRY_ALIAS: fsx-csi-driver
+          REPOSITORY: aws-fsx-csi-driver
```
wongma7, line 35: REPOSITORY doesnt appear to be used

wongma7:
/approve

lgtm thank you just 1 comment about the kustomize

deploy/kubernetes/overlays/stable/ecr-public/kustomization.yaml
wongma7: to be clear this is the diff I mean by "it should say..."
```
-  - ../../bases
+  - ../../../base
 images:
-  - name: 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/aws-fsx-csi-driver
-    newName: public.ecr.aws/fsx-csi-driver/aws-fsx-csi-driver
\ No newline at end of file
+  - name: amazon/aws-fsx-csi-driver:v0.9.0
+    newName: public.ecr.aws/fsx-csi-driver/aws-fsx-csi-driver
```

jacobwolfaws:
This is covered by this new PR: https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/291 
/close

PR #293, opened 2023-01-12
Have CSI driver use 2.12 Lustre client

jacobwolfaws:
update client
This PR moves the csi driver to use the 2.12 lustre client
**What testing is done?**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ RUN OS=$TARGETOS ARCH=$TARGETARCH make $TARGETOS/$TARGETARCH
 FROM amazonlinux:2 AS linux-amazon
 RUN yum update -y
 RUN yum install util-linux libyaml -y \
-    && amazon-linux-extras install -y lustre2.10
+    && amazon-linux-extras install -y lustre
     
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -22,7 +22,7 @@ RUN OS=$TARGETOS ARCH=$TARGETARCH make $TARGETOS/$TARGETARCH
 FROM amazonlinux:2 AS linux-amazon
 RUN yum update -y
 RUN yum install util-linux libyaml -y \
-    && amazon-linux-extras install -y lustre2.10
+    && amazon-linux-extras install -y lustre
```
l1n, line 25: Hey, question, why aren't we version pinning here anymore?

khoang98: We chose to move to a versionless repo to easier facilitate upgrades for customers. This helps protect against the scenario where customer pins an older version in some configuration file and forgets about it.

l1n: This seems like it risks issues with unintentional upgrades. Please, reconsider offering a pinned version here (or at least tag with the built Dockerfile version in the official ECR images)

Issue #305, opened 2023-04-04
Controller is being evicted every 5 minutes 


BEvgeniyS:
/kind bug

**What happened?**
Controller is being evicted every 5 minutes 

**What you expected to happen?**
If I set  
```
  nodeSelector:
    mynodetype: "true"

  tolerations:
  - effect: NoExecute
    key: mynodetype
    operator: Equal
    value: "false"
```
in values.yaml, I expect the controller to run on the node without being evicted every 5 minutes

**How to reproduce it (as minimally and precisely as possible)?**
- Have a node with taint:
  ```  - effect: NoExecute                                                                                                                                                                                     
    key: mynodetype                                                                                        
    value: "false"
  ```
- set the nodeSelector and toleration for the Controller in values.yaml:
  ```
  nodeSelector:
    mynodetype: "true"

  tolerations:
  - effect: NoExecute
    key: mynodetype
    operator: Equal
    value: "false"
  ```

**Anything else we need to know?**:
K8s will only apply first matched toleration. So the order of toleration matters, especially when there's `tolerationSeconds` involved. Universal timed `NoExecute` [toleration](https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/511427dbc5d931dcd03729d98755b61a601560c8/charts/aws-fsx-csi-driver/templates/controller-deployment.yaml#L40)
```
- operator: Exists
  effect: NoExecute
  tolerationSeconds: 300
```

 makes it impossible to meaningfully use other NoExecute tolerations because it's hardcoded to be on the top. Moving it to below configured values solves the issues for us.

**Environment**
- Kubernetes version (use `kubectl version`): v1.24.10-eks-48e63af
- Driver version: v0.8.3

jacobwolfaws:
@BEvgeniyS we've parametrized tolerations + released a new helm chart and I'm under the impression this solves the issues you were seeing. Going to close this, please let me know if another change needs to be made

PR #306, opened 2023-04-04
Parameterise tolerations and securityContext

BrunoChauvet:

Better control of `tolerations` is required when all nodes in a cluster have taints with effect `NoExecute`.
Fixes https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/305

This PR gives better control over `tolerations` so they can be entirely replaced with custom ones.
The same level of customisation is applied to `securityContext`.

Note that this follows the same conventions as `aws-ebs-csi-driver`
Controller: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/controller.yaml#L43
Daemonset: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/node.yaml#L39

**What testing is done?**

### DIFF
--- a/charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
@@ -35,14 +35,22 @@ spec:
       serviceAccountName: {{ .Values.controller.serviceAccount.name }}
       priorityClassName: system-cluster-critical
       tolerations:
-        - key: CriticalAddonsOnly
-          operator: Exists
-        - operator: Exists
-          effect: NoExecute
-          tolerationSeconds: 300
         {{- with .Values.controller.tolerations }}
         {{- toYaml . | nindent 8 }}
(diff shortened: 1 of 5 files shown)


jacobwolfaws:
also please increase the helm image version to 1.5.2 in the Chart.yaml & update the changelog, which will automatically release a new helm chart (and thus allow you to use the new helm chart). Here's an example of a recent PR I did: https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/304/commits/433d65311ea8762ab4063153c55e19147d551301

jacobwolfaws:
Tolerations were parametrized in this PR: https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/311. We don't currently support topology, please open a new PR to 1) add topology support or 2) parameterize securityContext. Thanks!

Issue #298, opened 2023-02-03
Update k8s version


dims:
Can we update to newer versions of k8s we vendor in? Are we sticking to an older version for a reason?

https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/master/go.mod#LL11C2-L11C2

jacobwolfaws:
We should be able to update to it with our next csi driver release, will resolve this issue when the new driver image is released.

jacobwolfaws:
k8s were bumped, closing

PR #309, opened 2023-05-11
Adopting Multiple CSI Driver improvements

jacobwolfaws:
new feature + QOL changes

This PR is about bring the FSx for Lustre CSI Driver closer in parity with the EBS CSI Driver: https://github.com/kubernetes-sigs/aws-ebs-csi-driver to improve the repo's best practices and allow the [start up taint change](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pulls) to be adopted within this driver as well. These changes include:
- Adding inflight checks to Create and Delete RPC's (best practice) (https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/702 https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/163)
- Adding feature gates to options (pre-emptively in preparation of json logging support) (https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1467)
- Modularize Node Service and Controller Service (preparation for start up taint pr & best practice) (https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/257)
- Increase MaxRetries (best practice https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/769)
- Improve metadata collection  (preparation for start up taint pr & best practice)
- Implementing operating modes  (preparation for start up taint pr & best practice) (https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/438) Note: I changed the implementation slightly so that customers need to explicitly set mode using the mode flag 
- update go dependencies and remove vendor dir (best practice and part of https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/298)
- Add json logging option (https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1467)
- Specify driver mode in the values.yaml for easier customization and more clear default recommendation

Note: will update e2e dependencies in future PR

**What testing is done?** 
- Manual testing (built driver image and tested dynamic provisioning + mounting a file system)
   - set json logging to confirms logged in json format properly
   - controller set with different mode values
   - confirmed that the driver image is built with the proper environment data

- sanity tests and unit tests
- e2e testing (in this PR)


pkg/cloud/metadata_k8s.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error getting Node %v: %w", nodeName, err)
+	}
+
+	providerID := node.Spec.ProviderID
+	if providerID == "" {
+		return nil, fmt.Errorf("node providerID empty, cannot parse")
+	}
+
+	awsInstanceIDRegex := "s\\.i-[a-z0-9]+|i-[a-z0-9]+$"
```
hughdanliu, line 47: nit: This regex retrieves the ec2 instance id of the node from the provider id, right? If so, I'm partial to specifying ec2InstanceId in the names of these variables. Do you have context into what would be required to provide support for the CSI driver in non-EKS Kubernetes environments?

Makefile
```diff
@@ -12,13 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-VERSION=v0.9.0
+VERSION?=v0.9.0
```
hughdanliu, line 15: nit: Is there more context behind this change? Is there any world in which VERSION would be already set before making this declaration?

charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
```diff
@@ -48,10 +48,19 @@ spec:
           image: {{ printf "%s:%s" .Values.image.repository (default (printf "v%s" .Chart.AppVersion) (toString .Values.image.tag)) }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
+            {{- if ne .Release.Name "kustomize" }}
+            - --mode=controller
+            {{- else }}
+            # - --mode={all,controller,node} # specify the driver mode
```
hughdanliu, line 54: Should this line be commented out? As is, if the release name is not "kustomize" then we set the mode to controller, if it is "kustomize" then we doing nothing. What is the default behavior if mode isn't set? Should we even give the option for users to specify a driver mode of "node" in the controller?

deploy/kubernetes/base/controller-deployment.yaml
```diff
@@ -34,10 +34,15 @@ spec:
           image: public.ecr.aws/fsx-csi-driver/aws-fsx-csi-driver:v0.9.0
           imagePullPolicy: IfNotPresent
           args:
+            # - --mode={all,controller,node} # specify the driver mode
```
hughdanliu, line 37: Same idea as above, for my own understanding what is the default behavior when mode isn't specified? From what I can tell the switch statements are configured to throw an error. Why don't we specify the mode as controller here?

.gitignore
```diff
@@ -6,3 +6,7 @@
 *.swp
 bin/
 *~
+
+# Vendor dir
+vendor/
```
hughdanliu, line 11: smart, we should do this for the OpenZFS CSI driver too

charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
jacobwolfaws: This was following precedent set by EBS, reviewing this with them to see if it should be removed

pkg/driver/driver.go
```diff
@@ -19,59 +19,51 @@ package driver
...
-		endpoint: endpoint,
-		cloud:    cloud,
-		inFlight: inFlight,
-		nodeID:   metadata.GetInstanceID(),
-		mounter:  newNodeMounter(),
+	driver := Driver{
+		options: &driverOptions,
 	}
+
+	driver.controllerService = newControllerService(&driverOptions)
```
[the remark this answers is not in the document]
jacobwolfaws: It's done like this to accommodate the different modes, the alternative would be to return a driver configured differently for each case. This is the format EBS uses + the one I prefer to that option

charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
jacobwolfaws: Added a new commit that moves mode specification to values.yaml and removed this.

cmd/options/node_options_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+func TestNodeOptions(t *testing.T) {
+	testCases := []struct {
+		name  string
+		flag  string
+		found bool
+	}{
+		{
+			name:  "lookup desired flag",
+			flag:  "volume-attach-limit",
```
torredil, line 33: Remove this.

pkg/driver/controller.go
```diff
@@ -59,7 +72,40 @@ const (
...
+
+// newControllerService creates a new controller service
+// it panics if failed to create the service
+func newControllerService(driverOptions *DriverOptions) controllerService {
+	region := os.Getenv("AWS_REGION")
+	if region == "" {
+		klog.V(5).InfoS("[Debug] Retrieving region from metadata service")
+		metadata, err := cloud.NewMetadataService(cloud.DefaultEC2MetadataClient, cloud.DefaultKubernetesAPIClient, region)
+		if err != nil {
+			klog.ErrorS(err, "Could not determine region from any metadata service. The region can be manually supplied via the AWS_REGION environment variable.")
```
torredil, line 90: missing `klog.FlushAndExit(klog.ExitFlushTimeout, 1)`

pkg/driver/driver.go
```diff
@@ -18,59 +18,76 @@ package driver
...
+	// ControllerMode is the mode that only starts the controller service.
+	ControllerMode = "controller"
+	// NodeMode is the mode that only starts the node service.
+	NodeMode = "node"
+	// AllMode is the mode that only starts both the controller and the node service.
+	AllMode = "all"
 )
 
-var (
-	volumeCaps = []csi.VolumeCapability_AccessMode{
```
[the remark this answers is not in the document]
jacobwolfaws: I moved it into controller.go, similar to how EBS does it: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/controller.go#L44

cmd/options/node_options_test.go
jacobwolfaws: good catch, removed + added ./cmd/... to test path in Makefile

pkg/driver/constants.go
```diff
@@ -0,0 +1,6 @@
+package driver
```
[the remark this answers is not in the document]
jacobwolfaws: We plan to add the startup taint key to constants.go in a separate PR.
Example from EBS CSI Driver: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/constants.go#L172-L175

Not opposed to removing constants.go, but I feel comfortable leaving it in right now

pkg/driver/controller.go
jacobwolfaws: this is currently the same behavior as EBS: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/controller.go#L88-L89 Should a similar change be made to the ebs csi driver?

pkg/driver/driver.go
```diff
@@ -93,8 +110,18 @@ func (d *Driver) Run() error {
...
+	case NodeMode:
+		csi.RegisterNodeServer(d.srv, d)
+	case AllMode:
+		csi.RegisterControllerServer(d.srv, d)
+		csi.RegisterNodeServer(d.srv, d)
+	default:
+		return fmt.Errorf("unknown mode: %s", d.options.mode)
+	}
 
 	klog.Infof("Listening for connections on address: %#v", listener.Addr())
```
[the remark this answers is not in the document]
jacobwolfaws: Reason: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#change-log-functions

I think this merits a separate commit/PR, because this change will need to be made in a lot of different places in the repo

pkg/driver/controller.go
```diff
@@ -71,10 +117,17 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Error(codes.InvalidArgument, "Volume capabilities not provided")
 	}
 
-	if !d.isValidVolumeCapabilities(volCaps) {
+	if !isValidVolumeCapabilities(volCaps) {
 		return nil, status.Error(codes.InvalidArgument, "Volume capabilities not supported")
```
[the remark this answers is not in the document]
jacobwolfaws: Sorry can you clarify what change you're recommending here?

pkg/driver/controller.go
```diff
@@ -71,10 +117,17 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+	// check if a request is already in-flight
+	if ok := d.inFlight.Insert(volName); !ok {
+		msg := fmt.Sprintf("Create volume request for %s is already in progress", volName)
+		return nil, status.Error(codes.Aborted, msg)
+	}
+	defer d.inFlight.Delete(volName)
+
 	// create a new volume with idempotency
 	// idempotency is handled by `CreateFileSystem`
 	volumeParams := req.GetParameters()
```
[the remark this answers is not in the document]
jacobwolfaws: I think this current format is a little easier to read + maintains consistency with the other drivers. Personal preference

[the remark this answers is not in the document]
jacobwolfaws: I think there's an opportunity to clean up code in the repo similar to this change, I'll add a TODO but think it's out of scope for this PR

pkg/driver/controller.go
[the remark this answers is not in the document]
jacobwolfaws: Yup that probably makes sense to do, I'll add a TODO for a future PR

pkg/driver/controller.go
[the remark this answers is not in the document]
hughdanliu: +1 to Eddie's comment. We followed this implementation pattern for the OpenZFS CSI driver. I'd also recommend throwing an error on the create in the default cause to be consistent with the FSx API - if the user provides an invalid parameter in the storage class, they shouldn't expect to have a successful creation.

pkg/driver/controller.go
torredil: panic also works here, I don't feel strongly about it.

jacobwolfaws: I'm going to add a note about this for more research, it's possible we should just move all panics to this

PR #312, opened 2023-05-22
Update dependencies & remove e2e vendor from repo

jacobwolfaws:

cleaning up the repo + updating dependencies
**What testing is done?** 
- sanity tests
- building image, deploying with helm, and testing dynamic provisioning


jacobwolfaws:
It seems like removing the e2e vendor causing issues with our e2e testing. We shouldn't let this block next image release

PR #317, opened 2023-05-23
Release 0.10 - part 3/3, merge to main

jacobwolfaws:

**What testing is done?**


jacobwolfaws:
Note: seems like it wouldn't go smoothly because of the failed container deployment, but passed checks + received lgtm so I'm merging it

PR #318, opened 2023-05-24
use minimal base csi driver image

jacobwolfaws:
N/A
- Use a minimal base image for the csi driver to reduce CVEs and reduce image size by 80% (according to our testing)
- Makes a change to our Makefile so now images are tagged/labeled with architecture prefixes rather than being <untagged>
**What testing is done?** 
- manually tested dynamic provisioning and mounting
- e2e tests

These changes are based off of the ebs csi driver [Makefile](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/Makefile) and [Dockerfile](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/Dockerfile)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-# Copyright 2019 The Kubernetes Authors.
+# Copyright 2023 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -58,20 +58,25 @@ bin/aws-fsx-csi-driver: | bin
...
-	docker buildx build \
-		--no-cache-filter=linux-amazon \
-		--platform=$(PLATFORM) \
-		--progress=plain \
-		--target=$(OS)-$(OSVERSION) \
-		--output=type=registry \
-		-t=$(IMAGE):$(TAG) \
-		.
-	touch $@
+all-push: all-image-registry push-manifest
```
PatrickGhadban, line 63: I think the `push-manifest` path leads to do everything `all-image-registry` in addition to also making the manifest. Is that expected? I don't think there's any harm in it but it might be worth testing having just `all-push: push-manifest `

PatrickGhadban: Alternatively - having `push-manifest` make the image and the manifest is a bit stretched for the name. You can change the path for `push-manifest` to not call `all-image-registry`. If not, could we change the name to something more encompassing like `all-image-manifest`? I know it's not a great name haha please feel free to come up with a better one

Makefile
```diff
@@ -58,20 +58,25 @@ bin/aws-fsx-csi-driver: | bin
 .PHONY: all
 all: all-image-docker
 
+# Builds all images and pushes them
```
PatrickGhadban, line 61: Nit: it builds the images *and manifest*. Could you change this to `Builds all images, and the manifest and push them to the public ECR`

jacobwolfaws:
transient error /retest

Makefile
jacobwolfaws: I think it's fine as is, image manifest seems to be implicit with images

Makefile
```diff
@@ -80,12 +85,13 @@ sub-image-%:
...
 .image-$(TAG)-$(OS)-$(ARCH)-$(OSVERSION):
 	docker buildx build \
-		--no-cache-filter=linux-amazon \
 		--platform=$(OS)/$(ARCH) \
 		--progress=plain \
 		--target=$(OS)-$(OSVERSION) \
 		--output=type=$(OUTPUT_TYPE) \
 		-t=$(IMAGE):$(TAG)-$(OS)-$(ARCH)-$(OSVERSION) \
+		--build-arg=GOPROXY=$(GOPROXY) \
+		--build-arg=VERSION=$(VERSION) \
```
khoang98, line 94: Doesn't having 2 args like this overwrite the first one?

jacobwolfaws: Nope it won't get overwritten, this is a supported behavior with docker build: https://docs.docker.com/engine/reference/commandline/build/#build-arg

PR #319, opened 2023-05-25
feat(pdb): add optional podDisruptionBudget to controller deployment

starlightromero:
Feature

Add optional podDisruptionBudget to controller deployment

**What testing is done?** 
Template locally podDistruptionBudget with and without values


jacobwolfaws:
Hi @starlightromero,

Thanks for the PR, I have some questions/thoughts about this:
1) Is there a reason why you didn't include the podDisruptionBudget annotation for the node pods? Curious to know why the controller pods were specifically targeted here.
2) Can you explain the issue/reason for this change (just for documentation purposes for when people look back at this PR)
3) How would you feel about changing pdb to podDisruptionBudget? pdb isn't immediately obvious on first glance (at least to me).
4) Can you rebase the two commits?

starlightromero:
1. podDisruptionBudget has been added for the controller pods as well as the node pods.
2. The `podDisruptionBudget` has been added in order to create a more stable EKS environment. The `podDisruptionBudget` allows a cluster administrator to define how a pod/replicaset should behave nodes go down or during an upgrade. There is the option to set `minAvailable` or `maxUnavailable`, which will determine the number of pods after the eviction.
3. `pdb` has been changes to `podDisruptionBudget` for clarity.
4. The commits have been rebased.

jacobwolfaws:
Hi @starlightromero, any thoughts on whether this check in the ebs csi driver equivalent is needed? https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/poddisruptionbudget-controller.yaml#LL1C1-L5C11 . Since this isn't being deployed by default I think it's okay as is.

Also it would make sense to update the `generate-kustomize` here: https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/master/Makefile#L131-L139 and run `make generate-kustomize` to update the deploy/* files accordingly. Thanks!

starlightromero:
Hi @jacobwolfaws, coming back to this PR.

For the check on the policy API version, it does not seem needed anymore. `policy/v1` has been available since K8s version `1.21`. I don't see the need to support clusters <= `1.20` but ultimately I'll let you make that call.

The `generate-kustomize` has been updated. Thanks for calling that out!

PR #321, opened 2023-05-26
Add region for controller to helm chart

jacobwolfaws:

Adding region to controller for easier specification in helm chart
**What testing is done?** 
- manually tested cases where region is set and region isn't set


torredil (approved):
/approve

PR #346, opened 2023-07-28
Upgrade k8s and eksctl in testing

jacobwolfaws:

Upgrading the k8s and eksctl versions for testing
**What testing is done?**


jacobwolfaws:
Not sure why the latest commit isn't being added to the PR, but just going to close and open a new one

PR #351, opened 2023-09-11
Add taint removal retry mechanism

jacobwolfaws:
both
PR is to give users a retry mechanism (if they so choose) for the csi driver. This allows users to address a noisy/overloaded control plane that can result in this error `connect: network is unreachable` when trying to remove the start up taint

**What testing is done?** 
Tested locally with
1) no retry mechanism
2) retry mechanism, passed on first attempt

to sanity check and make sure this doesn't impact nodes that don't have an error on first pass.

**Things to discuss**
1) flag name, is it clear and do we want something less wordy?
2) Advice on refactoring code/breaking out retry function so we can unit test

@torredil @ConnorJC3 since EBS may want to introduce a similar feature / port over this code once PR is merged

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -1,8 +1,9 @@
 # Driver Options
 There are a couple of driver options that can be passed as arguments when starting the driver container.
 
-| Option argument             | value sample                                      | default                                             | Description                                                                                 |
-|-----------------------------|---------------------------------------------------|-----------------------------------------------------|---------------------------------------------------------------------------------------------|
-| endpoint                    | tcp://127.0.0.1:10000/                            | unix:///var/lib/csi/sockets/pluginproxy/csi.sock    | The socket on which the driver will listen for CSI RPCs                                     |
-| extra-tags                  | key1=value1,key2=value2                           |                                                     | Tags specified in the controller spec are attached to each dynamically provisioned resource |
-| logging-format              | json                                              | text                                                | Sets the log format. Permitted formats: text, json                                          |
+| Option argument     | value sample            | default                                          | Description                                                                                      |
+|---------------------|-------------------------|--------------------------------------------------|--------------------------------------------------------------------------------------------------|
+| endpoint            | tcp://127.0.0.1:10000/  | unix:///var/lib/csi/sockets/pluginproxy/csi.sock | The socket on which the driver will listen for CSI RPCs                                          |
+| extra-tags          | key1=value1,key2=value2 |                                                  | Tags specified in the controller spec are attached to each dynamically provisioned resource      |
+| logging-format      | json                    | text                                             | Sets the log format. Permitted formats: text, json                                               |
+| retry-taint-removal | true                    | false                                            | If set to true, will keep retrying node taint removal for 90 seconds with an exponential backoff |
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -38,6 +38,7 @@ func main() {
 		driver.WithEndpoint(options.ServerOptions.Endpoint),
 		driver.WithMode(options.ServerOptions.DriverMode),
 		driver.WithExtraTags(options.ControllerOptions.ExtraTags),
+		driver.WithRetryTaintRemoval(options.NodeOptions.RetryTaintRemoval),
 	)
 
 	if err != nil {
--- a/cmd/options/node_options.go
+++ b/cmd/options/node_options.go
@@ -22,7 +22,9 @@ import (
 
 // NodeOptions contains options and configuration settings for the node service.
 type NodeOptions struct {
+	RetryTaintRemoval bool
 }
 
-func (o *NodeOptions) AddFlags(fs *flag.FlagSet) {
+func (s *NodeOptions) AddFlags(fs *flag.FlagSet) {
+	fs.BoolVar(&s.RetryTaintRemoval, "retry-taint-removal", false, "Max number of tries for node taint removal")
 }
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -52,9 +52,10 @@ type Driver struct {
 }
 
 type DriverOptions struct {
-	endpoint  string
-	mode      string
-	extraTags string
+	endpoint          string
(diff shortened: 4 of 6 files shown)


docs/options.md
```diff
@@ -1,8 +1,9 @@
...
+| Option argument         | value sample            | default                                          | Description                                                                                 |
+|-------------------------|-------------------------|--------------------------------------------------|---------------------------------------------------------------------------------------------|
+| endpoint                | tcp://127.0.0.1:10000/  | unix:///var/lib/csi/sockets/pluginproxy/csi.sock | The socket on which the driver will listen for CSI RPCs                                     |
+| extra-tags              | key1=value1,key2=value2 |                                                  | Tags specified in the controller spec are attached to each dynamically provisioned resource |
+| logging-format          | json                    | text                                             | Sets the log format. Permitted formats: text, json                                          |
+| taint-removal-num-tries | 1                       | 1                                                | Sets the maximum number of tries for removing the startup node taint.                       |
```
torredil, line 9: > Every configuration option represents a failure of the software to do the right thing automatically. Every configuration option needs to be documented and protected by unit tests, thereby increasing the cognitive load of user and developer alike. Sometimes they are necessary, but only as a last resort.  @wmesard

and here is another [gem](https://github.com/kubernetes/kubernetes/issues/65392#issuecomment-417213003):
> ...exposing the option to the end user is the easiest way out, but not necessarily the best. We should try to be smart and do the right thing automatically on behalf of the user.

This level of granularity is not necessary and users should not need to know internal implementation details such as `RemoveNotReadyTaintPollFactor = 1.8` etc to get a sense for how long 50 retries might take, or guess if X number of retries might fix their issue.

Ideally, this feature would be introduced with sensible defaults out of the box with no need for configuration. Users already go out of their way to apply the taint. If the configuration option is absolutely necessary, then lets consider a boolean (`enableRetry`), or time interval instead (`maxRetryDuration`).

torredil: Thanks for explaining the thought process @jacobwolfaws.

> there are non k8s users of csi drivers that aren't ever able to connect to k8s api. and thus would be stuck in whatever additional retries the default includes.

You correctly point out that this a K8s only feature. The above shouldn't be of concern because the taint removal feature is skipped if `CSI_NODE_NAME` is missing (set via downward API), which happens before we attempt to initialize the client:
https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/b02afc956ef7652c8de41a81c65858efae89fee7/pkg/driver/node.go#L297

Overall this is a good initiative, but wanted to emphasize that
1.  Introducing a configurable parameter is a one way door. It's much easier to add the option later than to remove it.
2. Adding resiliency against transient networking issues is typically something that should be managed on behalf of the user.

jacobwolfaws: > You correctly point out that this a K8s only feature. The above shouldn't be of concern because the taint removal feature is skipped if CSI_NODE_NAME is missing (set via downward API),

That's a good point I didn't consider. For future PR readers, here's the documentation on downward api:
https://kubernetes.io/docs/concepts/workloads/pods/downward-api/ .

Agreed with both your points, it makes more sense to introduce this as a default behavior and introduce a configurable parameter later on if this behavior has unexpected impact or needs greater configuration from users.

ConnorJC3: Here's how I see it, from the configuration perspective:

1. By default, we cannot be sure that we should have access to the k8s API server. Although the presence of `CSI_NODE_NAME` strongly implies we are running on Kubernetes, the cluster admin for example could be intentionally restricting access via a `NetworkPolicy`. We don't want to introduce a hard dependency on node pod access to the API server, so any retries here need to be non-intrusive and not significantly delay startup in the case where the API server truly isn't (and shouldn't be) available.

2. In a perfect world, where we knew with 100% certainty that the node had a taint applied, we would *never* want to give up retrying (or at the very least if we did, we would want to hard crash the pod).

3. Thus, my suggestion, if you want to introduce any configuration at all, would be to introduce a parameter along the lines of `--must-remove-node-taint`, that when set to `true` will cause the pod to hard fail when it is not able to reach the API server (and can also be used to extend the normal retry duration, as we know we are not wasting time in this case and we 100% do need to contact the API server).

torredil: > We don't want to introduce a hard dependency on node pod access to the API server, so any retries here need to be non-intrusive and not significantly delay startup in the case where the API server truly isn't (and shouldn't be) available.

This is a great point @ConnorJC3. Can we leverage API [responses](https://github.com/kubernetes/apimachinery/blob/a017454658b6eb672eb2d7b5eedad67f5f67665c/pkg/api/errors/errors.go#L659C1-L659C1) to handle these scenarios?

ConnorJC3: It is impossible to distinguish between the API server being intentionally inaccessible and temporary networking failures, as both will appear the same to the driver (without user configuration, as user input can 100% differentiate the two cases).

jacobwolfaws: So it seems to me the implementation should be:
1) have mustRemoveTaint (can work on name) default as false
2) leave the single remove node taint with soft failure as default
3) it doesn't seem like exponential backoff supports an indefinite number of retries, so I'll just set it to something extremely high.

jacobwolfaws: Talked offline, decided we don't want an infinite number of retries. Exponential backoff seems like the logical choice here considering the error users have seen is typically due to large scale ups.

kamirendawkins:
We have seen regular failures of startup taint removal, is there any progress on this feature? We feel it would be quite helpful as it does cause a blocking issue for other pods that are starting up and require fsx but are also trying to avoid the race between fsx and application readiness.

jacobwolfaws:
Hi @kamirendawkins thank you for this feedback! We plan to introduce this in the next csi driver release

jacobwolfaws:
/close
The ebs csi driver implemented this in a better way: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1861 going to create a new PR to implement background routine

PR #368, opened 2023-12-27
Support Data Repository Associations for Lusture 2.12 or newer filesystems(e.g. `PERSISTENT_2` deployment type)

### LINKED ISSUE #367: Supporting (multiple) data repository association
**Is your feature request related to a problem?/Why is this needed**

We would like to create `PERSISTENT_2` deployment type FSx volumes with (multiple) data repository associations via the kubernetes CSI. 

Currently, this driver only supports [CreateFileSystemLustreConfiguration.Auto{Import,Export}Path](https://docs.aws.amazon.com/fsx/latest/APIReference/API_CreateFileSystemLustreConfiguration.html) based configuration to link S3 bucket. But, this is not supported by `PERSISTENT_2` deployment type. `PERSISTENT_2` only supports [Data Repository Associations](https://docs.aws.amazon.com/fsx/latest/APIReference/API_DataRepositoryAssociation.html) to link S3 buckets.  Thus, we can currently create `PERSISTENT_2` filesystems with this driver but no buckets can be linked to the filesystems.

/feature

**Describe the solution you'd like in detail**

As describe in https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/252#issuecomment-1371774954:

```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: fsx-sc
provisioner: fsx.csi.aws.com
parameters:
  subnetId: <subnet-id>
  securityGroupIds: <security-group-id>
  deploymentType: PERSISTENT_2
  dataRepositories: |
    - fileSystemPath: /ns1/path1
      dataRepositoryPath: s3://s3-path-1
      BatchImportMetaDataOnCreate: true
    - fileSystemPath: /ns1/path2
      dataRepositoryPath: s3://s3-path-2
      BatchImportMetaDataOnCreate: true
```

**Additional context**

I'm a colleague with @ittus.  
I open this issue as instructed in https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/252#issuecomment-1381023146

everpeace:

new feature
fixes #367 

This PR supports [Data Repository Association](https://docs.aws.amazon.com/fsx/latest/LustreGuide/create-dra-linked-data-repo.html#create-linked-dra)([API reference](https://docs.aws.amazon.com/fsx/latest/APIReference/API_DataRepositoryAssociation.html)) for Lusture 2.12 or newer filesystems(e.g. `PERSISTENT_2` deployment type) like below:

```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: fsx-sc
provisioner: fsx.csi.aws.com
parameters:
  subnetId: subnet-0d7b5e117ad7b4961
  securityGroupIds: sg-05a37bfe01467059a
  deploymentType: PERSISTENT_2
  perUnitStorageThroughput: "125"
  # User can specify multiple data repository associations like this
  dataRepositoryAssociations: |
    - batchImportMetaDataOnCreate: true
      dataRepositoryPath: s3://ml-training-data-000
      fileSystemPath: /ml-training-data-000
      s3:
        autoExportPolicy:
          events: ["NEW", "CHANGED", "DELETED" ]
        autoImportPolicy:
          events: ["NEW", "CHANGED", "DELETED" ]
    - batchImportMetaDataOnCreate: true
      dataRepositoryPath: s3://ml-training-data-001
      fileSystemPath: /ml-training-data-001
      s3:
        autoExportPolicy:
          events: ["NEW", "CHANGED", "DELETED" ]
        autoImportPolicy:
          events: ["NEW", "CHANGED", "DELETED" ]

  # NOTE: These parameters can't be set when using　dataRepositoryAssociations
  #       as document explained:: https://docs.aws.amazon.com/fsx/latest/APIReference/API_CreateFileSystemLustreConfiguration.html
  # s3ImportPath: s3://ml-training-data-000
  # s3ExportPath: s3://ml-training-data-000/export
  # autoImportPolicy: NEW_CHANGED
```

**What testing is done?** 

```shell
make test

# with setting GINKGO_FOCUS=".*fsx-csi-e2e.*PERSISTENT_2.*"
make test-e2e 
```


charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
```diff
@@ -93,12 +93,14 @@ spec:
...
             - --csi-address=$(ADDRESS)
             - --v={{ .Values.sidecars.provisioner.logLevel }}
-            - --timeout=5m
+            - --timeout=$(TIMEOUT)
             - --extra-create-metadata
             - --leader-election=true
           env:
             - name: ADDRESS
               value: /var/lib/csi/sockets/pluginproxy/csi.sock
+            - name: TIMEOUT
```
jacobwolfaws, line 102: This should be part of a separate commit

pkg/cloud/cloud.go
```diff
@@ -62,6 +64,10 @@ var (
 	// disks are found with the same volume name.
 	ErrMultiFileSystems = errors.New("Multiple filesystems with same ID")
 
+	// ErrMultiAssociations is an error that is returned when multiple
+	// associations are found with the same volume name.
```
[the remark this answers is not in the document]
everpeace: Thanks. fixed in 04efd399daf20fe7b4d85df9e61059591c54bebe.

charts/aws-fsx-csi-driver/templates/controller-deployment.yaml
everpeace: I made this change in separate commit: https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/368/commits/22c8a714ac7fa12b8a5e9f2175a11121b035a826

Sorry for force push 🙇

everpeace (commented):
@jacobwolfaws I addressed your feedbacks. PTAL 🙇

charts/aws-fsx-csi-driver/values.yaml
```diff
@@ -65,6 +65,7 @@ controller:
     - effect: NoExecute
       operator: Exists
       tolerationSeconds: 300
+  provisionerTimeout: 5m
```
everpeace, line 68: Should we make the default provisioner timeout longer in helm chart? It is because it often takes more time to prepare an FSx filesystem when it has data repository associations.

Single FSx for Lusture fielsystem can have [up to 8 data repository associations](https://arc.net/l/quote/fccabakk). 

In my experience, it usually takes around 7-10 minutes to make single data repository associations available even for empty S3 bucket.

So, I think `90 min = 10x8 min (data repository associations) + 5 min (FSx filesystem) + <buffer>` would be safe because the current `CreateVolume` operation is synchronous operation and not be safe when timeout happened.

What do you think??

jacobwolfaws: I think keeping the default timer the same + clearly documenting the need to change the timeout if using DRAs would be the correct move. This ensures consistent behavior for users who aren't using DRAs. Extending it is a one way door (because reducing the timeout would break compatibility for users who are using a large number of DRAs).

everpeace: > Extending it is a one way door (because reducing the timeout would break compatibility for users who are using a large number of DRAs)

It makes sense.

> the default timer the same + clearly documenting the need to change the timeout if using DRAs would be the correct move

OK. let me add the documentation.

everpeace: > Extending it is a one way door (because reducing the timeout would break compatibility for users who are using a large number of DRAs)

It makes sense.

> the default timer the same + clearly documenting the need to change the timeout if using DRAs would be the correct move

OK. let me add the documentation.

everpeace: addressed in below commits:
- 9ee7b8f7f192e32d312a8bbb0c72fab137ddaec9
- 8d23e8130358f58949cab2dedb9209dd5cbbe6e6

deploy/kubernetes/base/controller.env
```diff
@@ -0,0 +1 @@
+CONTROLLER_PROVISIONER_TIMEOUT=5m
```
jacobwolfaws, line 1: What's the value of creating a separate file for this vs. putting it in the values.yaml:
https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/master/charts/aws-fsx-csi-driver/values.yaml#L42-L67

everpeace: This file is for manifests only for kustomize. `values.yaml` is dedicated to helm chart. I understand this driver supports both kustomize and helm.

In kustomize, injecting parameter in kustomize needs a bit hack. This env file is needed for kustomize users to change timeout value. I also updated `install.md` as below:

https://github.com/everpeace/aws-fsx-csi-driver/blob/suppor-dra/docs/install.md#deploy-driver
```
# To set CSI controller's provisioner timeout,
# Please follow the instruction
$ cd $(mktemp -d)
$ kustomize init
$ kustomize edit add resource "github.com/kubernetes-sigs/aws-fsx-csi-driver/deploy/kubernetes/overlays/stable/?ref=release-1.1"
$ kustomize edit add configmap fsx-csi-controller --from-literal=CONTROLLER_PROVISIONER_TIMEOUT=30m --behavior=merge
$ kubectl apply -k .
```

charts/aws-fsx-csi-driver/values.yaml
jacobwolfaws: I'm not sure about the information for users, it seems like users using DRAs will still be fine in most cases:
https://github.com/kubernetes-csi/external-provisioner?tab=readme-ov-file
https://github.com/kubernetes-csi/external-provisioner?tab=readme-ov-file#csi-error-and-timeout-handling
The CreateVolume will timeout and other ones will be made with an exponential backoff. It's only in the case of a large number of DRAs where this will be an issue.

deploy/kubernetes/base/controller.env
jacobwolfaws: I think we should avoid hacks when possible and this seems like an avoidable instance. If users want to configure their kustomize templates, they can download them, configure them, and deploy them freely. We should follow precedent in terms of implementation, which is to put it in the values.yaml.

pkg/cloud/cloud.go
```diff
@@ -41,11 +41,13 @@ const (
...
-	// for a completed create/update operation. FSx for Lustre filesystem
-	// creation time is around 5 minutes, and update time varies depending on
+	// PollCheckTimeout specifies the time limit for polling
+	// DescribeFileSystems/DescribeDataRepositoryAssociation for a completed
+	// create/update operation. FSx for Lustre filesystem
+	// creation time is around 5 minutes, Data Repository Association
+	// creation time is around 10 minutes, and update time varies depending on
 	// target file system values
-	PollCheckTimeout = 10 * time.Minute
+	PollCheckTimeout = 15 * time.Minute
```
jacobwolfaws, line 50: If PollCheckTimeout < provisionerTimeout, the provisionerTimeout will always kill the CreateVolume call before it the PollCheckTimeout is hit. I don't think incrementing this should make a difference

jacobwolfaws:
This has been up for a while, sorry. Going to freeze this PR for now, seems like there are some open comments and design decisions to be made here
/lifecycle-frozen

jacobwolfaws:
Seems like I can't freeze a PR :(

/remove-lifecycle rotten

PR #371, opened 2024-01-25
chore: Bump aws-sdk-go for EKS Pod Identity support

askulkarni2:
Bumps up aws-sdk-go to v1.50.3.

AWS SDK [v.1.47.11](https://github.com/aws/aws-sdk-go/releases/tag/v1.47.11) updated credential provider logic support which is required by pods which need to use the newly launched [Amazon EKS Pod Identities](https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-eks-pod-identity/).

Also see https://docs.aws.amazon.com/eks/latest/userguide/pod-id-minimum-sdk.html for further information.

**What testing is done?** 

Ran `make` and `make test` with no errors.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module sigs.k8s.io/aws-fsx-csi-driver
 
 require (
-	github.com/aws/aws-sdk-go v1.47.4
+	github.com/aws/aws-sdk-go v1.50.3
 	github.com/container-storage-interface/spec v1.9.0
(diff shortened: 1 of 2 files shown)


jacobwolfaws:
Hi @askulkarni2 we're going to be bumping all dependencies (not just aws sdk go) prior to our next release, but we're currently waiting on some other PRs to land before we do so.

/close

PR #373, opened 2024-02-16
Fix taint removal retry for non-swallowed errors

jacobwolfaws:
bug fix
Node taint removal should retry on unknown errors rather than fail

credits to @ConnorJC3 for the EBS PR for this: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1898
**What testing is done?** 
- e2e & sanity
- ran in dev cluster to confirm taint removal works as expected


torredil (approved):
/lgtm
/hold

PR #377, opened 2024-02-22
Add Affinity to DaemonSet

### LINKED ISSUE #376: Add Affinity to DaemonSet
**Is your feature request related to a problem? Please describe.**
If using the chart in a cluster with fargate enabled and several nodes with different taints, the daemonset tries to deploy the fsx-plugin also to those fargate nodes, which is not possible. 

**Describe the solution you'd like in detail**
Adding a configurable affinity definition to the helm chart would enable one to specify an antiaffinity for fargate

karaktaka:
Feature: Fixes #376 

See Issue #376 

**What testing is done?** 
This chart is currently deployed in our cluster and is working fine (besides, it's a very small change)

### DIFF
--- a/charts/aws-fsx-csi-driver/templates/node-daemonset.yaml
+++ b/charts/aws-fsx-csi-driver/templates/node-daemonset.yaml
@@ -46,6 +46,9 @@ spec:
         - key: "fsx.csi.aws.com/agent-not-ready"
           operator: "Exists"
         {{- end }}
+      {{- with .Values.node.affinity }}
+      affinity: {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
         - name: fsx-plugin
           securityContext:
--- a/charts/aws-fsx-csi-driver/values.yaml
+++ b/charts/aws-fsx-csi-driver/values.yaml
@@ -93,6 +93,15 @@ node:
     - operator: Exists
       effect: NoExecute
       tolerationSeconds: 300
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+        - matchExpressions:
+          - key: eks.amazonaws.com/compute-type
+            operator: NotIn
+            values:
+            - fargate
 
 nameOverride: ""
 fullnameOverride: ""


jacobwolfaws:
Hi @karaktaka, can you rebase the commits into a single commit?

karaktaka:
> Hi @karaktaka, can you rebase the commits into a single commit?

Hey @jacobwolfaws,
certainly. Does this work for you?

jacobwolfaws:
Sorry for the delayed response,

It might be worth adding something like this from the ebs-csi-driver: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml#L336-L344 where this is automatically set for users. Would like to hear your thoughts on that. Otherwise, this looks good to me.

karaktaka:
> Sorry for the delayed response,
> 
> I think it would make sense to add something like this from the ebs-csi-driver: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml#L336-L344 where this is automatically set for users, but affinity is still configurable. Would like to hear your thoughts on that. Otherwise, this looks good to me.

I just added the values from the ebs chart as I think they make sense like that.

jacobwolfaws:
> > Sorry for the delayed response,
> > I think it would make sense to add something like this from the ebs-csi-driver: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml#L336-L344 where this is automatically set for users, but affinity is still configurable. Would like to hear your thoughts on that. Otherwise, this looks good to me.
> 
> I just added the values from the ebs chart as I think they make sense like that.

No worries, sorry I should have been more clear. This restriction doesn't apply to our driver:
```
          - key: node.kubernetes.io/instance-type
            operator: NotIn
            values:
            - a1.medium
            - a1.large
            - a1.xlarge
            - a1.2xlarge
            - a1.4xlarge
```
also can you rebase the commits into a single commit? Other than these two things looks good imo. Will look at why the e2e test is failing for you

charts/aws-fsx-csi-driver/templates/node-daemonset.yaml
```diff
@@ -46,6 +46,10 @@ spec:
         - key: "fsx.csi.aws.com/agent-not-ready"
           operator: "Exists"
         {{- end }}
+      {{- with .Values.node.affinity }}
+      affinity:
+      {{- toYaml . | nindent 8 }}
```
jacobwolfaws, line 51: nit: this could be inline with affinity

jacobwolfaws:
@karaktaka sorry for taking so long on this, it seems like I don't get notifications when changes are force-pushed 🤔 . One tiny nit 😅 and I've put up a PR for pinning to a new AMI.

charts/aws-fsx-csi-driver/templates/node-daemonset.yaml
karaktaka: There you go

torredil (commented):
/lgtm

jacobwolfaws:
tested locally and behavior appears to work as expected.
Thanks @karaktaka for your contribution to the driver!
/lgtm
/approve

toneill818:
Can a new release of the helm chart be created with this change?

Issue #382, opened 2024-07-01
Allow configurable timeout for WaitForFileSystemAvailable 


jon-rei:
**Is your feature request related to a problem? Please describe.**

We are starting to use Scratch 2 FSx FileSystems and are doing so successfully for small ones. But when we try to create FSx volumes >50TB we always run into timeouts:

```
E0701 17:57:05.334673       1 driver.go:104] "GRPC error" err="rpc error: code = Aborted desc = Create volume request for pvc-1234 is already in progress"
E0701 17:57:06.262188       1 driver.go:104] "GRPC error" err=<
	rpc error: code = Internal desc = Filesystem is not ready: RequestCanceled: request context canceled
	caused by: context canceled
 >
```

It seems that the timeout is always 5 minutes, but creating a 72TB volume for example takes ~12 minutes.

**Describe the solution you'd like in detail**

It would be great if this timeout could be increased or made configurable since it's hardcoded [here](https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/master/pkg/cloud/cloud.go#L48).

**Describe alternatives you've considered**

Right now we are trying to create the filesystems manually and then create a PV + PVC. But we could really create the volumes dynamically for the pipelines we want to run.

jon-rei:
I wrote a small script to check the creation time of the filesystems and found that as they get bigger they take longer to create. I did my tests in eu-west-1 and I'm not sure if this is something specific to that region.

```
FSx for Lustre file system with 4800 GiB was available after: 6 minutes 44 seconds
FSx for Lustre file system with 9600 GiB was available after 9 minutes 6 seconds
FSx for Lustre file system with 24000 GiB was available after 8 minutes 32 seconds
FSx for Lustre file system with 50400 GiB was available after 9 minutes 5 seconds
FSx for Lustre file system with 100800 GiB was available after 11 minutes 19 seconds
```

So I think it would be great if the timeout could either grow dynamically with storage capacity or be available for manual configuration.

jon-rei:
@jonded94 and I dug a little deeper to figure out where the timeout was coming from and found that it's actually happening on the `DescribeFileSystemsWithContext` call, probably around this [line](https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/master/pkg/cloud/cloud.go#L383).
We think this is coming from the GO sdk package.

jacobwolfaws:
Might be related to
[...]

PR #384, opened 2024-07-04
Make Volume creation timeout configurable

jonded94:

Bug fix

Hopefully solves Issue: https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/382

Made volume creation timeout configurable by introducing a new Volume parameter "pollTimeout". This is a timeout in seconds that will be used on creation of the volume.

Found no way easily right now to make a poll timeout during resizing of the volume configurable.

**What testing is done?** 

None unfortunately, I don't have an easy way of testing this.


jonded94:
This is an alternative way to the PR https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/383. I personally don't care about a certain approach, it's just that currently, this driver unfortunately blocks us from creating large (>~60-70TiB) volumes and would like to solve that as soon as possible :slightly_smiling_face:

Issue #386, opened 2024-08-15
Create New Helm Chart Release


toneill818:
**Is your feature request related to a problem? Please describe.**
Create a new version of the helm chart that includes the updates since 2.9.0

**Describe the solution you'd like in detail**
Chart version 2.9.1 gets created

**Describe alternatives you've considered**
Downloading the source code and pointing helm to the currently installed

jacobwolfaws:
Might be easiest to just cut 2.9.2, job for 2.9.1 broken since it's using the old releaser: https://github.com/kubernetes-sigs/aws-fsx-csi-driver/actions/runs/10599639792/workflow

PR #387, opened 2024-08-15
chore: Bump chart version

toneill818:
Bump chart version which will include PR #377 

Creates a new release with the changes.
Issue #386 
**What testing is done?**

### DIFF
--- a/charts/aws-fsx-csi-driver/CHANGELOG.md
+++ b/charts/aws-fsx-csi-driver/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Helm chart
 
+# v1.9.1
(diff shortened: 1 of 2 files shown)


jacobwolfaws:
@toneill818 you need to do your CLA Authorization.

/ok-to-test

toneill818:
It looks like the e2e test failed due to an account limit in AWS. There may be an issue with cleaning up the fsx shares in that account unless it is used by other projects too. 

> failed to created filesystem CreateFileSystem failed: ServiceLimitExceeded: Account 209411653980 can create up to 100800 GiB of deployment type SCRATCH_1

Issue #255, opened 2022-09-13
Data repository integration s3 automatic sync


ilkarataev:
Hi gyus, Thank you for you work.
I see that Fsx Lustre now can do full bi-directional synchronization for s3. 
We need configure an automatic export policy for your data repository association, files on your FSx for Lustre file system are automatically exported to your data repository.
In storage class we don't have this option.
![image](https://user-images.githubusercontent.com/27670867/189835225-f528875e-4cf2-4c82-9631-5d8a8089e35d.png)
I see that in api aws 2. Have AutoExportPolicy and AutoImportPolicy
https://github.com/aws/aws-sdk-go/blob/v1.44.96/service/fsx/api.go#L15954-L15964
https://docs.aws.amazon.com/sdk-for-go/api/service/fsx/#AutoImportPolicy
Thank you and have a nice day.

PatrickGhadban:
Hi, thank you for bringing this to our attention. I've created a ticket for us to add support for AutoExportPolicy to dynamic provisioning. In the meantime, it is possible to create a Lustre file system with AutoExportPolicy & AutoImportPolicy and use static provisioning then it could be mounted inside container as a volume using the Driver.

markBETA:
Hi!

I am also interested in this feature. Is there a timeline for when we can expect it to be available? With the latest release, I saw that you have migrated to AWS SDK v2, which supports the AutoExportPolicy parameter.

Thank you!

PR #397, opened 2024-10-20
Migrate to aws sdk v2

### LINKED ISSUE #385: switch to aws-sdk-go-v2 as v1 is going away
Context: https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/
Migration guide: https://aws.github.io/aws-sdk-go-v2/docs/migrating/

Basically maintenance mode on July 31, 2024 and reach end-of-support on July 31, 2025. The SDK will not 
receive API updates for new or existing services, or be updated to support new regions.

michael-diggin:
Neither, changes all uses of `aws-sdk-go` to `aws-sdk-go-v2` - resolves #385.

As mentioned in #385, aws-sdk-go-v1 won't be receiving any more updates. Migrating to v2 allows for using newer features from FSx (eg AutoExportPolicy as mentioned in #255, this is something I need and is my motivation for this PR).

**What testing is done?** 
`make test`

I've aimed to make as few changes to the interfaces as possible, hoping it's just a 'drop-in' replacement, following https://aws.github.io/aws-sdk-go-v2/docs/migrating/
***Summary of changes***
1. `aws-sdk-go` -> `aws-sdk-go-v2` and added `aws-sdk-go-v2/config`, `aws-sdk-go-v2/service/fsx`, `aws-sdk-go-v2/feature/ec2/imds` to go.mod.
2. Session is replaced by Config
3. FSx types moved to `aws-sdk-go-v2/service/fsx/types`, and enums are now used rather than strings in many of the structs.
4. Many FSx fields were changed from `int64` to `int32`. (This did mean I've changed the types of some of the fields in some of the public APIs eg `pkg/cloud/cloud.go FileSystem`, alternatively they could be kept as int64 and the uses could be converted to int32 where needed.)
5. `ec2metadata` is replaced by `ec2/imds`
6. `ec2/imds` in v2 doesn't have an `Available()` function, however in v1 all that did was check `GetMetadata`, so I've implemented `Available()` here following the same logic.


pkg/cloud/cloud.go
```diff
@@ -129,16 +126,16 @@ type cloud struct {
 // NewCloud returns a new instance of AWS cloud
 // It panics if session is invalid
 func NewCloud(region string) (Cloud, error) {
-	awsConfig := &aws.Config{
-		Region:                        aws.String(region),
-		CredentialsChainVerboseErrors: aws.Bool(true),
```
jacobwolfaws, line 134: Is there a `CredentialsChainVerboseErrors` equivalent in v2?

pkg/cloud/cloud.go
```diff
@@ -129,16 +126,16 @@ type cloud struct {
...
+	awsConfig := aws.Config{
+		Region: region,
+		// Set RetryMaxAttempts to a high value. It will be "overwritten" if context deadline comes sooner.
+		RetryMaxAttempts: 8,
 	}
 
 	os.Setenv("AWS_EXECUTION_ENV", "aws-fsx-csi-driver-"+driverVersion)
 
-	svc := fsx.New(session.Must(session.NewSession(awsConfig)))
+	//svc := fsx.New(session.Must(session.NewSession(awsConfig)))
```
jacobwolfaws: nit: should remove comment

jacobwolfaws:
Hi @michael-diggin, Thanks for posting this PR! Can you squash the commits into a single commit?
/ok-to-test

pkg/cloud/cloud.go
michael-diggin: I couldn't find one, but could find https://github.com/aws/aws-sdk-go-v2/issues/98 so I don't think there is an alternative at this stage.

pkg/cloud/cloud.go
michael-diggin: Removed in a later commit.

pkg/cloud/cloud.go
jacobwolfaws: hm, yeah it doesn't seem like there's one. Thanks for looking into this

michael-diggin:
> Hi @michael-diggin, Thanks for posting this PR! Can you squash the commits into a single commit? /ok-to-test

Hey @jacobwolfaws, thanks for the quick review! I've squashed the two commits.

jacobwolfaws:
/approved
Since this is a rather large PR, going to find a second reviewer to this, but this seems good to me. My largest concern was any changes being made to user experience in terms of yaml files, but not seeing anything here. Thanks again!

jacobwolfaws:
/approve
Since this is a rather large PR, going to find a second reviewer to this, but this seems good to me. My largest concern was any changes being made to user experience in terms of yaml files, but not seeing anything here. Thanks again!

pkg/util/util.go
```diff
@@ -90,8 +91,8 @@ func ParseEndpoint(endpoint string) (string, string, error) {
 	return scheme, addr, nil
 }
 
-func roundUpSize(volumeSizeBytes int64, allocationUnitBytes int64) int64 {
-	return (volumeSizeBytes + allocationUnitBytes - 1) / allocationUnitBytes
+func roundUpSize(volumeSizeBytes int64, allocationUnitBytes int64) int32 {
+	return int32((volumeSizeBytes + allocationUnitBytes - 1) / allocationUnitBytes)
```
torredil, line 95: I think we can silently overflow with large inputs here.

Casting in several different places is rather confusing and increases the likelihood of bugs. If the expected output from this function is `int32`, can we make sure that the inputs are also `int32`?

pkg/cloud/metadata_ec2.go
```diff
@@ -1,23 +1,45 @@
...
+//
+// This is taken from v1 of aws-sdk-go to keep the interface the same.
+// See https://github.com/aws/aws-sdk-go/blob/main/aws/ec2metadata/api.go
+func (i *imdsClient) Available() bool {
+	input := &imds.GetMetadataInput{Path: "instance-id"}
+	if _, err := i.GetMetadata(context.Background(), input); err != nil {
+		return false
+	}
+	return true
+}
```
torredil, line 30: To avoid duplicating this code block, can we use the exported function here https://github.com/aws/aws-sdk-go/blob/7112c0a0c2d01713a9db2d57f0e5722225baf5b5/aws/ec2metadata/api.go#L208 instead?

torredil (commented):
PR largely lgtm, thank you for the contribution.

pkg/cloud/metadata_ec2.go
michael-diggin: I'm not sure we should, the PR is trying to move away from using that Go library and use the v2 version.
In the v2 version, this `Available()` that gets used by the aws-fsx-csi-driver no longer exists, but all it does it precisely this check on `GetMetadata`, so I've implemented it here as a smaller blast radius change, as it is part of an interface that imdsClient needs to implement. A better change could be to change the uses of `Available` in this repo to something else but that felt like it could add too many things to this PR.

michael-diggin:
> PR largely lgtm, thank you for the contribution.

Thanks for the review, I've replied to your comments and made a small update. Let me know what you think when you get another chance to review it.

pkg/cloud/metadata_ec2.go
torredil: Hmm, I see. That makes sense, in that case, I suggest getting rid of `Available` altogether. As I understand it, this validation is completely unnecessary. For example, if EC2 IMDS is not available, this will return an error: https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/79364b4c9952b28ddabe228c3eed6a01b7792338/pkg/cloud/metadata.go#L80

In other words, lets attempt to retrieve the relevant information from IMDS without the redundant check to see if it is available first with `EC2MetadataInstanceInfo` -- if that returns an error, then we fall back to retrieving the information from K8s.

michael-diggin: Makes sense, I've just pushed that change and updated the tests. Several of the tests were checking for specific errors from the `EC2MetadataInstanceInfo` call, and as now it's not returning that error I've updated those tests to return a k8s client error like one of the already existing tests does.

pkg/util/util.go
torredil: ack - Thanks for letting us know about that pain point. I generally agree that we should make the minimal set of changes for the migration and not blow up the scope of this PR; maintainers of this repo ought to be aware of this and help with cleaning up the casting situation to prevent regressions in the future.

TLDR - CSI spec expects and returns values in `int64`, for example:
```
int64 volume_capacity_bytes = 2;
```

but the SDK's expected input/output might be of type`int32` for some attributes.

In other words, we need to work towards a state where minimal casting is done - preferably, all inputs are in `int32` where needed, then a single cast to `int64` is performed in a centralized place. That should be safe because `int32` is _implicitly_  convertible to `int64`- every 32 bit int can be expressed as a 64 bit int, so there is no risk of silent overflows, etc.

cc: @jacobwolfaws

pkg/cloud/metadata_ec2.go
torredil: Great, thanks for the quick turnaround 🙂

torredil:
/approve

Code looks good for migration. @michael-diggin @jacobwolfaws please manually test the scenario where IMDS is not available to validate that there are no regressions in expected behavior before this is released. The driver should successfully fall back to retrieving the needful information from k8s API without crashing.

pkg/util/util.go
michael-diggin: Makes sense, thanks for the extra detail!
Just thinking, we could leave everything in `RoundUpVolumeSize` in int64, there's only two uses of that function (https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/master/pkg/driver/controller.go#L213 and https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/master/pkg/driver/controller.go#L395). I could change it there such that they cast to int32 only in those locations, and even check if the number to cast doesn't exceed `Math.MaxInt32`, with an error if it is?

jacobwolfaws:
Hi @michael-diggin @torredil, just catching up to this thread.
@michael-diggin 
can you squash these into a single commit (or two if you feel like there's a natural separation)
Will try to test with imds disabled in the next couple of days so that we can get this merged, thanks for all your work!

michael-diggin:
> @michael-diggin can you squash these into a single commit (or two if you feel like there's a natural separation) Will try to test with imds disabled in the next couple of days so that we can get this merged, thanks for all your work!

Done, thanks for offering to test it!

jacobwolfaws:
Confirmed that this falls back to kubernetes api successfully when imds is disabled (also tested dynamic provisioning)
```
I1031 14:14:00.407477       1 driver.go:61] "Driver Information" Driver="fsx.csi.aws.com" Version="vsdk2"
I1031 14:14:00.407582       1 node.go:60] "[Debug] Retrieving node info from metadata service"
I1031 14:14:00.407602       1 metadata.go:72] "retrieving instance data from ec2 metadata"
I1031 14:14:03.533898       1 metadata.go:75] "ec2 metadata is not available"
I1031 14:14:03.533921       1 metadata.go:83] "retrieving instance data from kubernetes api"
I1031 14:14:03.534376       1 metadata.go:88] "kubernetes api is available"
I1031 14:14:03.541869       1 node.go:66] "regionFromSession Node service" region=""
I1031 14:14:03.543952       1 mount_linux.go:282] Detected umount with safe 'not mounted' behavior
I1031 14:14:03.544330       1 driver.go:127] "Listening for connections" address="/csi/csi.sock"
I1031 14:14:03.547718       1 node.go:347] "No taints to remove on node, skipping taint removal"
I1031 14:14:04.837509       1 node.go:243] "NodeGetInfo: called" args={}
```
/lgtm

PR #402, opened 2025-01-06
Add dynamic provisioning support for metadata configuration and efa

jpayvazian:
New feature

Adds dynamic provisioning support for `PERSISTENT_2` Lustre filesystems with metadata configuration and EFA enabled ones.

Example storage class:
```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: fsx-sc
provisioner: fsx.csi.aws.com
parameters:
  subnetId: <subnet-id>
  securityGroupIds: <sg-id>
  deploymentType: PERSISTENT_2
  perUnitStorageThroughput: "1000"
  fileSystemTypeVersion: "2.15"
  metadataConfigurationMode: "AUTOMATIC"
  efaEnabled: "true"
mountOptions:
  - flock
```

**What testing is done?** 
- make test
- manually deployed driver image and validated filesystem creation for AUTOMATIC/USER_PROVISIONED modes and efaEnabled


pkg/driver/controller_test.go
```diff
@@ -289,6 +290,90 @@ func TestCreateVolume(t *testing.T) {
 				mockCtl.Finish()
 			},
 		},
+		{
+			name: "success: normal with deploymentType PERSISTENT_2 and metadataConfiguration",
```
jacobwolfaws, line 294: I notice you didn't add a test in controller_test for efaEnabled. Is that intentional?

jacobwolfaws:
Correct me if I'm wrong, but work needs to be done in terms of configuring EKS nodes for EFA to run as expected with the cluster. Can the instructions be added as something like `efa_instructions.md` and referenced in the dynamic_provisioning readme

pkg/cloud/cloud.go
```diff
@@ -149,6 +152,7 @@ func (c *cloud) CreateFileSystem(ctx context.Context, volumeName string, fileSys
 	}
 
 	lustreConfiguration := &types.CreateFileSystemLustreConfiguration{}
+	metadataConfiguration := &types.CreateFileSystemLustreMetadataConfiguration{}
```
jacobwolfaws, line 155: nit: seems like this could be defined in the
```
if fileSystemOptions.MetadataConfigurationMode != "" {
...
}
```
section since it's only used in this instance

pkg/cloud/cloud.go
```diff
@@ -193,6 +197,17 @@ func (c *cloud) CreateFileSystem(ctx context.Context, volumeName string, fileSys
 		lustreConfiguration.WeeklyMaintenanceStartTime = aws.String(fileSystemOptions.WeeklyMaintenanceStartTime)
 	}
 
+	if fileSystemOptions.EfaEnabled {
+		lustreConfiguration.EfaEnabled = aws.Bool(true)
+	}
+
+	if fileSystemOptions.MetadataConfigurationMode != "" {
+		metadataConfiguration.Mode = types.MetadataConfigurationMode(fileSystemOptions.MetadataConfigurationMode)
+		if fileSystemOptions.MetadataIops != 0 {
```
jacobwolfaws, line 206: Is there a reason why we are specifically checking if it's `!= 0` here? Shouldn't we just be validating that it's an int? As it stands, it seems like this could lead to a confusing user experience.

jacobwolfaws: Can you test to see what happens if you set it to 0?

ayvaziaj: setting to 0:
```
CreateFileSystem failed: operation error FSx: CreateFileSystem, https response error StatusCode: 400, RequestID: ef933f08-10e1-4ea0-b968-9dadcc5981d3, BadRequest: Must specify metadata IOPS in USER_PROVISIONED mode
```
vs setting to 1:
```
BadRequest: 1 validation error detected: Value at 'lustreConfiguration.metadataConfiguration.iops' failed to satisfy constraint: Member must have value greater than or equal to 1500
```

The latter api response seems better if set to 0 so I can update this check, I think its fine to leave for the other parameters like perUnitStorageThroughput since this is the response if setting that to 0.

```
CreateFileSystem, https response error StatusCode: 400, RequestID: bdbc7c71-2031-420c-bc63-714653651205, InvalidPerUnitStorageThroughput: Invalid Lustre deployment configuration: The storage throughput (MB/s/TiB) value should be one of [125,250,500,1000]
```

PR #403, opened 2025-01-13
Bump dependencies for v1.3.0

jpayvazian:
Bumping dependencies for new release
```
go get -u ./...
go mod vendor
go mod tidy
```

Added unimplemented fields/fixed mocks for `mock_k8s_client.go` to resolve errors similar to 
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/mock_k8s_client.go

**What testing is done?** 
make, make-test


jacobwolfaws:
Seems like changes need to be made similar to this commit: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/commit/528276fb3b1ed4865d6a31b5dfaa5a93ba618577

PR #422, opened 2025-05-02
Update provenance check for changed message

kuba-wolf:
Bug fix
In the provenance check we were checking using this pattern `*"See 'docker buildx build --help'."*` but the message was updated to be `Run 'docker buildx build --help' for more information`. Just reducing the specificity of the check should be sufficient

**What testing is done?** 
Ran builld from personal, not seeing an error message from provenance check anymore


kuba-wolf:
/hold
seems like there's still an issue. Testing additional changes

PR #430, opened 2025-05-15
Node region

coolstim:

**What testing is done?**


kuba-wolf:
Hi @coolstim, thanks for adding this, I actually just noticed it during testing. I'm not going to be able to give the final ship but can ok to test.

Separately, can you:
1) squash the commits
2) show some logs from the node pod with this change to show that it is working as expected.

I can't think of a reason ec2 metadaservice is actually needed by our node csi pod, long term that should probably be refactored out. That shouldn't block this fix though

/ok-to-test

khoang98:
Taking a look

khoang98:
Approved the PR. This will be rolled into our next release

Issue #433, opened 2025-05-29
Filesystem is not ready: operation error FSx: DescribeFileSystems, context canceled"


rileyhun:
/kind bug

**What happened?**

We are using FSx CSI Driver to dynamically provision FSx for Lustre file systems. The CSI Driver is faulty and keeps continuously creating FSx for Lustre file systems until the subnet runs out of IPs.

This is the error we are getting:

```
E0529 21:42:10.691980       1 driver.go:105] "GRPC error" err="rpc error: code = Internal desc = Filesystem is not ready: operation error FSx: DescribeFileSystems, context canceled"
I0529 21:43:12.994422       1 cloud.go:351] "WaitForFileSystemAvailable" filesystem="fs-084c19b0f9eb5b7f2" status="CREATING"
I0529 21:43:42.995196       1 cloud.go:351] "WaitForFileSystemAvailable" filesystem="fs-084c19b0f9eb5b7f2" status="CREATING"
I0529 21:44:13.058396       1 cloud.go:351] "WaitForFileSystemAvailable" filesystem="fs-084c19b0f9eb5b7f2" status="CREATING"
I0529 21:44:42.995050       1 cloud.go:351] "WaitForFileSystemAvailable" filesystem="fs-084c19b0f9eb5b7f2" status="CREATING"
I0529 21:45:12.979509       1 cloud.go:351] "WaitForFileSystemAvailable" filesystem="fs-084c19b0f9eb5b7f2" status="CREATING"
I0529 21:45:43.145222       1 cloud.go:351] "WaitForFileSystemAvailable" filesystem="fs-084c19b0f9eb5b7f2" status="CREATING"
I0529 21:46:13.001566       1 cloud.go:351] "WaitForFileSystemAvailable" filesystem="fs-084c19b0f9eb5b7f2" status="CREATING"
I0529 21:46:43.028694       1 cloud.go:351] "WaitForFileSystemAvailable" filesystem="fs-084c19b0f9eb5b7f2" status="CREATING"
I0529 21:47:13.100871       1 cloud.go:351] "WaitForFileSystemAvailable" filesystem="fs-084c19b0f9eb5b7f2" status="CREATING"
E0529 21:47:42.760804       1 driver.go:105] "GRPC error" err="rpc error: code = Internal desc = Filesystem is not ready: operation error FSx: DescribeFileSystems, context canceled"
```

While the file system does get created successfully, there are some errors in the CSI Driver logs like this:

```
[...]

PR #435, opened 2025-06-05
Allow the timeout of the csi driver to be set

coolstim:
This allows the configuration of a previously hard coded setting that controls the timeout of call to the csi driver.
https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/433

AWS had an issue taking the provisioning of file systems longer than the 5m default. Now this setting is configurable.

**What testing is done?** 
Deployment with a timeout of 15m during the time that provisioning was taking longer than 5min.


samuelverstraete (approved):

coolstim:
@khoang98 or @torredil any chance to get a review here?

PR #453, opened 2025-08-19
Fix CSI Driver creating multiple filesystems per PVC

Trickybrain:
Bug fix.

Fix duplicate filesystem creation in CSI driver

**Issue**

The CSI driver creates duplicate filesystems when the initial filesystem creation takes longer than 5 minutes. This happens because:

CSI provisioner times out after 5 minutes and retries
The retry uses the same ClientRequestToken, but AWS's idempotency window has expired
This results in multiple filesystems being created for the same PVC

**CloudTrail evidence:**

Before fix: Multiple CreateFileSystem API calls with same ClientRequestToken (~5 minutes apart)
After fix: Single CreateFileSystem API call, subsequent calls use DescribeFileSystems

**Fix**

Added filesystem lookup by volume name before creation. If filesystem exists:

Skip creation
Wait for filesystem to be available
Return existing filesystem
This prevents duplicate creation regardless of timeout or token expiry.

**Testing**

Verified in CloudTrail: single CreateFileSystem call per volume
Successfully creates and reuses existing filesystem
Handles CSI provisioner timeout without creating duplicates


pkg/cloud/cloud.go
```diff
@@ -445,3 +446,71 @@ func isBadRequestUpdateInProgress(err error) bool {
...
+		output, err := c.fsx.DescribeFileSystems(ctx, input)
+		if err != nil {
+			return nil, fmt.Errorf("failed to describe filesystems: %v", err)
+		}
+
+		klog.V(5).InfoS("Checking batch of filesystems", "count", len(output.FileSystems))
+
+		// Search current batch
+		for _, fs := range output.FileSystems {
+			// Skip if filesystem is being deleted
```
khoang98, line 471: nit: Let's modify this so that if the lifecycle is not AVAILABLE or CREATING, continue

pkg/cloud/cloud.go
```diff
@@ -445,3 +446,71 @@ func isBadRequestUpdateInProgress(err error) bool {
...
+		}
+
+		// Check if more results exist
+		if output.NextToken == nil {
+			break
+		}
+		nextToken = output.NextToken
+	}
+
+	klog.V(4).InfoS("No existing filesystem found", "volumeName", volumeName)
```
khoang98, line 514: Make this V(2)

pkg/cloud/cloud.go
```diff
@@ -445,3 +446,71 @@ func isBadRequestUpdateInProgress(err error) bool {
...
 }
+
+func (c *cloud) FindFileSystemByVolumeName(ctx context.Context, volumeName string) (*FileSystem, error) {
+	var nextToken *string
+	const maxResults = 100
+
+	klog.V(4).InfoS("Searching for existing filesystem", "volumeName", volumeName)
+
+	for {
+		input := &fsx.DescribeFileSystemsInput{
```
khoang98, line 457: nit: I'd leave a comment explaining that we are doing this pagination/looping since DescribeFileSystems does not support filtering by tags

khoang98:
Change PR name to be a bit more specific

PR #460, opened 2025-09-12
Add Lustre client init container for AL userspace tool

khoang98:
A combination of the two.

Amazon-Linux based distros have kernel module and userspace tool separate. CSI driver users on AL are missing the userspace on the underlying node level, which is needed for server to propagate different configurations to the client. 

**What testing is done?** 
Manual.


olemarkus (approved):

khoang98:
/hold
Might want to add a lever for disabling this and a check if userspace already installed

deploy/kubernetes/base/node-daemonset.yaml
```diff
@@ -35,6 +35,25 @@ spec:
                 values:
                 - fargate
                 - hybrid
+      initContainers:
+      - name: lustre-client-installer
+        image: amazonlinux:2
```
candonov, line 40:
```suggestion
        image: public.ecr.aws/amazonlinux/amazonlinux:2023
```

PR #494, opened 2026-04-17
Pin github actions to commit SHAs

khoang98:
Fix

Fix breaking actions

**What testing is done?** 
NA


MichaelKwok1998:
lgtm

Issue #492, opened 2026-04-15
Support Driver Metrics in FSx CSI Driver


vardaro:
**Is your feature request related to a problem?/Why is this needed**

Netflix is using this CSI Driver to mount FSx Lustre on our K8s platform. The lack of OOTB Prometheus metrics on provisioning lifecycle presents an operational gap for cluster administrators. We'd like this driver to support driver metrics (similar to what is supported in EBS CSI Driver https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/71266f21dc6e0ff8e58bba7058de8bbb44fd91b7/docs/metrics.md?plain=1#L8). I'm happy to drive the change personally if we can gain consensus on the feature here

/feature

**Describe the solution you'd like in detail**

Similar to the EBS CSI Driver, we'd like a way to enable driver metrics and see information related to AWS API calls, and on-instance storage utilization metrics (observed IOPS/Throughput).

**Describe alternatives you've considered**

Cluster administrators can collect metrics outside of the driver via a DaemonSet. We do something like this for online page cache instrumentation. We'd rather delegate metrics submission to the driver, if possible.

**Additional context**

khoang98:
Thanks for opening this PR @vardaro and for the offer to drive the implementation! 

Here are my thoughts - let me know what you think:

I think it makes sense to have a PR that adds Prometheus metrics for FSx API calls (request duration, errors, throttles) on the controller side, similar to what the EBS CSI driver does. The CSI sidecar metrics (provisioner, resizer) can also be enabled as part of this work with minimal effort. This covers the provisioning lifecycle observability gap you described.  We can use the EBS CSI driver's pkg/metrics/ is a good reference for the implementation pattern. The main differences would be the metric namespace (aws_fsx_csi_ vs aws_ebs_csi_) and the smaller set of API operations to instrument (CreateFileSystem, DeleteFileSystem, DescribeFileSystems, UpdateFileSystem).

The on-instance storage utilization metrics I'm more unsure about. The EBS driver's NVMe metrics work because EBS volumes are block devices with a vendor-specific NVMe log page that exposes per-volume performance stats via ioctl. For Lustre, there's no NVMe device or equivalent per-volume performance counter interface. The best I can think of here is to use the Lustre client-side llite stats (exposed via /proc) but there are a few issues w/ this:
[...]

PR #497, opened 2026-05-07
Add support for API metrics in the FSx CSI Driver

vardaro:

New feature

Per https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/492 this PR adds support for FSx API CSI Driver prometheus metrics. This helps driver operators gain better clarity on failure and monitor request latencies. As discussed in the Github issue, I tried to model the structure and implementation of this to closely track the [EBS CSI Driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver).

I added these ones to start:

| Metric name | Metric type | Description | Labels |
|-------------|-------------|-------------|--------|
| `aws_fsx_csi_api_request_duration_seconds` | Histogram | AWS SDK API request duration by request type in seconds | `request=<FSx API operation name>` |
| `aws_fsx_csi_api_request_errors_total` | Counter | Total number of AWS SDK API errors by error code and request type | `request=<FSx API operation name>`, `code=<error code>` |
| `aws_fsx_csi_api_request_throttles_total` | Counter | Total number of throttled AWS SDK API requests per request type | `request=<FSx API operation name>` |

This PR is just the driver metrics. We plan to follow up with the sidecar metrics in a different PR to scope this one down.

**What testing is done?** 

I tested this internally on Netflix's K8s platform.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -97,9 +97,10 @@ func (m *MetricRecorder) InitializeMetricsHandler(address, path, certFile, keyFi
 	mux.Handle(path, rateLimitMiddleware(limiter, metricsHandler))
 
 	server := &http.Server{
-		Addr:        address,
-		Handler:     mux,
-		ReadTimeout: 3 * time.Second,
+		Addr:         address,
+		Handler:      mux,
+		ReadTimeout:  3 * time.Second,
+		WriteTimeout: 5 * time.Second,
 	}
 
 	go func() {
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -47,7 +47,7 @@ var (
 
 // MetricRecorder holds a Prometheus registry and the registered metrics.
 type MetricRecorder struct {
-	mu       sync.Mutex
+	mu       sync.RWMutex
 	registry *prometheus.Registry
 	metrics  map[string]any
 }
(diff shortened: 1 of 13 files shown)


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,216 @@
...
+	once.Do(func() {
+		r = &MetricRecorder{
+			registry: prometheus.NewRegistry(),
+			metrics:  make(map[string]any),
+		}
+	})
+	return r
+}
+
+func (m *MetricRecorder) initializeMetricWithOperations(name, help string, labelNames []string) {
```
khoang98, line 71: Do we want to pre-register counters here? Not just the histogram?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,216 @@
...
+		"CreateFileSystem",
+		"DeleteFileSystem",
+		"DescribeFileSystems",
+		"UpdateFileSystem",
+	}
+)
+
+// MetricRecorder holds a Prometheus registry and the registered metrics.
+type MetricRecorder struct {
+	mu       sync.Mutex
```
khoang98: We have generally fewer API calls compared to EBS, but it could be slightly less contentious to use a RWMutex here

khoang98:
@vardaro This looks really good - a few minor comments from me. Let me know what you think!

pkg/metrics/metrics.go
vardaro: I considered it! `APIRequestThrottles` could be pre-seeded (it only has the request label), but `APIRequestErrors` can't(?) be fully pre-seeded due to the dynamic `code` label. The EBS driver leaves both counters lazily registered, so I opted to follow that pattern for consistency.

pkg/metrics/metrics.go
vardaro: Nice catch. I moved this to `sync.RWMutex`, makes more sense to me to be consistent w/ EBS.

khoang98:
@vardaro apologies I've just come back from a vacation - I will review this early next week so we can get this merged

khoang98:
/approved
/lgtm

We'll roll this into the next release, thank you for your work!

Issue #495, opened 2026-04-24
NodeUnpublishVolume hangs indefinitely when umount(2) blocks, permanently holding the operation lock


jspiewak:
### Describe the bug

When a pod using an FSx for Lustre volume is deleted, the CSI node plugin calls `umount(2)` on
the volume's target path as part of `NodeUnpublishVolume`. If the `umount(2)` syscall blocks
(e.g., due to a temporarily unresponsive FSx endpoint or an in-progress I/O flush that never
completes), the goroutine executing `NodeUnpublishVolume` never returns. Because the driver
acquires a per-volume operation lock before beginning the unmount, and holds it for the duration,
that lock is never released.

All subsequent `NodePublishVolume`, `NodeUnpublishVolume`, `NodeStageVolume`, and
`NodeUnstageVolume` calls on the same volume fail immediately with:

```
operation for volume <volume-id> is already in progress
```

The result is that the original pod stays in `Terminating` forever. On our cluster we have
observed pods stuck in this state for **44+ days**. There is no self-healing path: the driver has
no lock timeout, and kubelet has no way to interrupt the in-flight syscall. The only recovery is
to restart the CSI node DaemonSet pod on the affected node, which clears the in-memory lock
(requiring `--force --grace-period=0` if the CSI node pod itself is also stuck).

### Steps to reproduce / how we detected it

We do not have a reliable reproduction recipe, but the failure is detectable through the following
observable state:

1. Pod is in `Terminating` with a deletion timestamp set weeks or months ago.
2. Describing the pod shows it is waiting on a CSI volume to be unmounted.
3. Logs from the `fsx-csi-node` pod on the same node contain a stream of entries like:
   ```
   An error occurred in NodeUnpublishVolume, operation for volume <id> is already in progress
   ```
4. No corresponding "operation complete" log entry exists — the original operation never finished.
5. Restarting the `fsx-csi-node` pod (force-delete if also stuck) unblocks the termination
   immediately.

### Environment

- **aws-fsx-csi-driver version**: v1.8.0 (also present in v1.9.0 based on changelog review)
- **Helm chart version**: 1.15.0
- **Kubernetes**: Amazon EKS
- **FSx filesystem type**: FSx for Lustre

### Root cause

The `NodeUnpublishVolume` gRPC handler acquires a per-volume lock, then calls `mounter.Unmount()`
which wraps the `umount(2)` syscall. The gRPC context (which carries the kubelet-imposed deadline)
[...]

PR #499, opened 2026-06-24
Add Feature for Forceful unmount when Initial unmount hangs 

MichaelKwok1998:
A new feature
Add a feature for forceful unmount to prevent hanging unmount command. 
**What testing is done?**

### DIFF
--- a/charts/aws-fsx-csi-driver/templates/node-daemonset.yaml
+++ b/charts/aws-fsx-csi-driver/templates/node-daemonset.yaml
@@ -68,6 +68,12 @@ spec:
             - --endpoint=$(CSI_ENDPOINT)
             - --logging-format={{ .Values.node.loggingFormat }}
             - --v={{ .Values.node.logLevel }}
+            {{- with .Values.node.forcefulUnmountTimeout }}
+            - --forceful-unmount-timeout={{ . }}
+            {{- end }}
+            {{- with .Values.node.extraArgs }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
           env:
             - name: CSI_ENDPOINT
               value: unix:/csi/csi.sock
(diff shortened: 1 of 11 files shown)


jspiewak:
Some data from a live wedge on v1.9.0 that bears on this PR — full details and the stack trace in #495.

**The approach still works, but for a different reason than the one stated.** We captured the block in `os/exec.(*Cmd).Wait` → `pidfdWait` → `Waitid`, reaping a `umount` child stuck in `ll_kill_super`, not in the `CombinedOutput` pipe read. Running the unmount in a goroutine and returning once the mount is gone is therefore still the right shape — it just isn't rescuing a pipe that never reaches EOF, it's declining to wait on a child that never exits.

**Two questions on the poll condition.** At wedge time the mount was **already absent** from the host mount table, and in an earlier occurrence the target directory itself had been removed while `umount` was still blocked.

1. If `target` no longer exists, `IsLikelyNotMountPoint` returns an error from `os.Stat`, so `if err == nil && notMnt` never fires and the poll runs to `forcefulUnmountTimeout` instead of succeeding early. Is that intended? Treating "path is gone" as success would recover this case within one poll interval.
2. Would the `umount -f` escalation actually help here? The syscall is blocked inside the Lustre client with the mount already detached, so I would expect a forced unmount to block the same way — meaning the timeout path may not recover even in cases where the poll path would.

**Enablement.** Two things mean merging this will not reach Helm users as-is: `--forceful-unmount-timeout` defaults to `-1` (feature off), and `charts/aws-fsx-csi-driver/templates/node-daemonset.yaml` hardcodes the node container args with no `extraArgs` hook, so there is no supported way to set the flag. Is chart plumbing in scope for this PR or a follow-up?

khoang98:
@jspiewak Thanks for the data! Think most of your points should be addressed in Michael's latest push. For the case where the client is stuck in `ll_kill_super` - you're right that `umount -f` likely won't help. However I think it's still worth having where the mount is genuinely stuck rather than already detached.

PR #500, opened 2026-07-08
Use idempotency token + FS cache for createVolume

khoang98:
Bug fix

We can reduce the number of API calls needed to populate the in-memory volume cache by having a FS cache. 

**What testing is done?** 
```
I0707 21:24:53.299402       1 controller.go:120] "CreateVolume: called" args="name:\"pvc-de547017-3d8b-4567-a3c4-8a2c26b0047f\"  capacity_range:{required_bytes:1288490188800}  volume_capabilities:{mount:{}  access_mode:{mode:MULTI_NODE_MULTI_WRITER}}  parameters:{key:\"csi.storage.k8s.io/pv/name\"  value:\"pvc-de547017-3d8b-4567-a3c4-8a2c26b0047f\"}  parameters:{key:\"csi.storage.k8s.io/pvc/name\"  value:\"test-idempotency\"}  parameters:{key:\"csi.storage.k8s.io/pvc/namespace\"  value:\"default\"}  parameters:{key:\"deploymentType\"  value:\"SCRATCH_2\"}  parameters:{key:\"securityGroupIds\"  value:\"sg-0b2278d56973e7658\"}  parameters:{key:\"subnetId\"  value:\"subnet-0d3e6d835c32210de\"}"
I0707 21:24:54.455940       1 cloud.go:283] "CreateFileSystem: cached filesystem ID" volumeName="pvc-de547017-3d8b-4567-a3c4-8a2c26b0047f" fileSystemId="fs-01440f8d72faa3bfa"
I0707 21:25:25.052960       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:25:54.737242       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:26:24.787188       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:26:54.938189       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:27:25.286816       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:27:54.709627       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:28:24.728749       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:28:54.817791       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:29:24.925148       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:29:54.300466       1 controller.go:120] "CreateVolume: called" args="name:\"pvc-de547017-3d8b-4567-a3c4-8a2c26b0047f\"  capacity_range:{required_bytes:1288490188800}  volume_capabilities:{mount:{}  access_mode:{mode:MULTI_NODE_MULTI_WRITER}}  parameters:{key:\"csi.storage.k8s.io/pv/name\"  value:\"pvc-de547017-3d8b-4567-a3c4-8a2c26b0047f\"}  parameters:{key:\"csi.storage.k8s.io/pvc/name\"  value:\"test-idempotency\"}  parameters:{key:\"csi.storage.k8s.io/pvc/namespace\"  value:\"default\"}  parameters:{key:\"deploymentType\"  value:\"SCRATCH_2\"}  parameters:{key:\"securityGroupIds\"  value:\"sg-0cc85cf7574986971,sg-0b2278d56973e7658\"}  parameters:{key:\"subnetId\"  value:\"subnet-0d3e6d835c32210de\"}"
E0707 21:29:54.300704       1 driver.go:105] "GRPC error" err="rpc error: code = Aborted desc = Create volume request for pvc-de547017-3d8b-4567-a3c4-8a2c26b0047f is already in progress"
I0707 21:29:54.456398       1 inflight.go:73] "Volume operation finished" key="pvc-de547017-3d8b-4567-a3c4-8a2c26b0047f"
E0707 21:29:54.456419       1 driver.go:105] "GRPC error" err="rpc error: code = Internal desc = Filesystem is not ready: operation error FSx: DescribeFileSystems, context canceled"
I0707 21:29:56.302212       1 controller.go:120] "CreateVolume: called" args="name:\"pvc-de547017-3d8b-4567-a3c4-8a2c26b0047f\"  capacity_range:{required_bytes:1288490188800}  volume_capabilities:{mount:{}  access_mode:{mode:MULTI_NODE_MULTI_WRITER}}  parameters:{key:\"csi.storage.k8s.io/pv/name\"  value:\"pvc-de547017-3d8b-4567-a3c4-8a2c26b0047f\"}  parameters:{key:\"csi.storage.k8s.io/pvc/name\"  value:\"test-idempotency\"}  parameters:{key:\"csi.storage.k8s.io/pvc/namespace\"  value:\"default\"}  parameters:{key:\"deploymentType\"  value:\"SCRATCH_2\"}  parameters:{key:\"securityGroupIds\"  value:\"sg-0cc85cf7574986971,sg-0b2278d56973e7658\"}  parameters:{key:\"subnetId\"  value:\"subnet-0d3e6d835c32210de\"}"
I0707 21:29:56.302336       1 cloud.go:479] "GetFileSystemIdByVolumeName: cache hit" volumeName="pvc-de547017-3d8b-4567-a3c4-8a2c26b0047f" fileSystemId="fs-01440f8d72faa3bfa"
I0707 21:29:56.302358       1 controller.go:147] "CreateVolume: found cached filesystem ID, skipping CreateFileSystem" volumeName="pvc-de547017-3d8b-4567-a3c4-8a2c26b0047f" fsId="fs-01440f8d72faa3bfa"
I0707 21:30:27.307049       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:30:57.287146       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:31:27.358202       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:31:57.208759       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="CREATING"
I0707 21:32:27.337756       1 cloud.go:375] "WaitForFileSystemAvailable" filesystem="fs-01440f8d72faa3bfa" status="AVAILABLE"
I0707 21:32:27.337789       1 inflight.go:73] "Volume operation finished" key="pvc-de547017-3d8b-4567-a3c4-8a2c26b0047f"
```

### DIFF
--- a/hack/e2e/cleanup-old-filesystems.sh
+++ b/hack/e2e/cleanup-old-filesystems.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Cleans up FSx Lustre filesystems older than a specified age.
+# This prevents stranded filesystems from e2e test failures from accumulating
+# and exhausting the account's FSx storage quota.
+
+set -euo pipefail
+
+REGION=${AWS_REGION:-us-west-2}
+MAX_AGE_DAYS=${MAX_AGE_DAYS:-7}
+
+echo "Cleaning up FSx Lustre filesystems older than ${MAX_AGE_DAYS} days in ${REGION}"
+
+CUTOFF_EPOCH=$(( $(date +%s) - (MAX_AGE_DAYS * 86400) ))
+
+# Get all Lustre filesystem IDs and creation times
+FS_LIST=$(aws fsx describe-file-systems \
+  --region "${REGION}" \
+  --output json \
+  --query 'FileSystems[?FileSystemType==`LUSTRE`].{Id:FileSystemId,Created:CreationTime,Capacity:StorageCapacity,Lifecycle:Lifecycle}')
+
+if [[ -z "${FS_LIST}" || "${FS_LIST}" == "[]" || "${FS_LIST}" == "null" ]]; then
+  echo "No Lustre filesystems found"
+  exit 0
+fi
+
+TOTAL=$(echo "${FS_LIST}" | jq length)
+echo "Found ${TOTAL} Lustre filesystem(s)"
+
+DELETED=0
+SKIPPED=0
+
+for row in $(echo "${FS_LIST}" | jq -r '.[] | @base64'); do
+  _jq() {
+    echo "${row}" | base64 --decode | jq -r "${1}"
+  }
+
+  FS_ID=$(_jq '.Id')
+  CREATION_TIME=$(_jq '.Created')
+  CAPACITY=$(_jq '.Capacity')
+  LIFECYCLE=$(_jq '.Lifecycle')
+
+  # FSx API returns creation time as Unix epoch float (e.g., 1754330438.291)
+  # Truncate to integer for comparison
+  FS_EPOCH=${CREATION_TIME%.*}
+
+  if [[ "${FS_EPOCH}" -lt "${CUTOFF_EPOCH}" ]]; then
+    AGE_DAYS=$(( ($(date +%s) - FS_EPOCH) / 86400 ))
+    echo "  DELETE ${FS_ID}: ${AGE_DAYS} days old, ${CAPACITY} GiB, lifecycle=${LIFECYCLE}"
+
+    if [[ "${LIFECYCLE}" == "DELETING" ]]; then
+      echo "    Already deleting, skipping"
+      SKIPPED=$((SKIPPED + 1))
+      continue
+    fi
+
+    aws fsx delete-file-system \
+      --region "${REGION}" \
+      --file-system-id "${FS_ID}" \
+      --output text > /dev/null 2>&1 || echo "    WARNING: failed to delete ${FS_ID}"
+
+    DELETED=$((DELETED + 1))
+  else
+    SKIPPED=$((SKIPPED + 1))
+  fi
+done
+
+echo "Cleanup complete. Deleted: ${DELETED}, Skipped: ${SKIPPED}"
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -18,7 +18,6 @@ package driver
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"strconv"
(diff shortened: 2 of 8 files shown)


khoang98:
I caused a unit test failure with my original changed - that's fixed. Now, the test acct is hitting its FSx quota limit. I added in a cleanup script to nuke FSxL file systems older than 7 days for this acct. Letting that test now

hack/e2e/cleanup-old-filesystems.sh
```diff
@@ -0,0 +1,82 @@
...
+# Cleans up FSx Lustre filesystems older than a specified age.
+# This prevents stranded filesystems from e2e test failures from accumulating
+# and exhausting the account's FSx storage quota.
+
+set -euo pipefail
+
+REGION=${AWS_REGION:-us-west-2}
+MAX_AGE_DAYS=${MAX_AGE_DAYS:-7}
+
+echo "Cleaning up FSx Lustre filesystems older than ${MAX_AGE_DAYS} days in ${REGION}"
```
torredil, line 26: It would be better to add cleanup support for FSx resources in the [boskos AWS janitor](https://github.com/kubernetes-sigs/boskos/tree/master/aws-janitor/resources) instead. Cleanup would then run independently of presubmits and use the janitor’s existing TTL and tag filtering. If we do it here, then for example stranded filesystems are only cleaned when the CI job in this repo runs.

pkg/driver/controller.go
```diff
@@ -140,43 +139,45 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
...
 	defer d.inFlight.Delete(volName)
 
-	existingFS, err := d.cloud.FindFileSystemByVolumeName(ctx, volName)
-	if err != nil && !errors.Is(err, cloud.ErrNotFound) {
-		return nil, status.Errorf(codes.Internal, "Failed to check existing filesystem: %v", err)
-	}
-
+	// Check if we already know the filesystem ID for this volume (from a previous CreateFileSystem call).
+	// If so, skip CreateFileSystem and go straight to polling DescribeFileSystems.
+	// This avoids calling CreateFileSystem with a potentially expired ClientRequestToken on retry.
```
torredil, line 144: Since we have agreed to use `ClientRequestToken` for idempotency, `CreateVolume` should simply call `CreateFileSystem` with `req.Name` as the idempotency token. That is sufficient, and what I'd recommend. Idempotency should be fully handled server-side, and the CSI driver's responsibility here is simply to pass the token to the create API.

If we need protection beyond the token lifetime, then we should implement find-or-create against the backend, similar to [EBS CreateSnapshot](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/903e2a13bb5c934edc693a0ee00f1c5603ed1251/pkg/driver/controller.go#L836) (_unfortunately, we have to adopt that strategy there because CreateSnapshot does not let you pass in an idempotency token at all, but importantly, please notice how there is no cache at all, its not needed_). You just look up the resource by its tag and return it when found, otherwise create it with the tag in the same request.

I do not recommend adding a local cache at all (ie remove `fsIdCache`), it does not protect us across restarts or controller pod leader election changes and it can hide bugs until a cold cache retry.

torredil (changes requested):

hack/e2e/cleanup-old-filesystems.sh
khoang98: Yeah I think longer term moving cleanup to that makes sense. In short term I'd like to just add this here to clean up since I think our e2e is stranding somewhere.

pkg/driver/controller.go
khoang98: Yeah understood that this doesn't protect across restarts - but I do think it has some value here. The cache is just attempt to mitigate the risk of token expiration (though this is extremely slim chance). And understood that it doesn't survive restart or crashes - it's just here as a slight improvement to just the token alone.

Let me know what you think.

torredil: Okay, I will just disagree and commit on keeping the cache. 

AIUI, it introduces more risk than value and opens the door to bad behavior. For example, right now the ID is cached before the fs becomes available, but the entry is never removed if creation reaches FAILED or the filesystem is deleted out of band. Every subsequent retry then continues using the same ID and returns the same error. That is one concrete example, and I worry there are other edge cases in this unnecessary code path that will be difficult to catch, even with more code coverage.

I think this is a bad idea, but I'm willing to take the gamble with you.

hack/e2e/cleanup-old-filesystems.sh
torredil: boskos AWS janitor PR: https://github.com/kubernetes-sigs/boskos/pull/260

pkg/driver/controller.go
khoang98: That makes sense to me. I think we have to go with this approach since we already got alignment from a few stakeholders here - and if we see issues in the future, we can revisit

